PR #33, opened 2023-04-19
Minor improvements

slang25:
Some minor improvements while browsing the code. The code base seems to have mixed line endings, so sorry for the messy diff, which every line endings I go for it's a mess 😄 

[Review with whitespace off](https://github.com/awslabs/aws-dotnet-messaging/pull/33/files?w=1)

It might be best to view commit-by-commit too.


normj:
Congrats on being the first external contributor! 

Is the change to use pascal case in logging string format a style preference or am I missing some significance about the casing?

src/AWS.Messaging/Configuration/PublisherTargetType.cs
```diff
@@ -1,23 +1,23 @@
...
+    /// </summary>
+    public const string SQS_PUBLISHER = "SQS";
+    /// <summary>
+    /// Specifies a publisher that will publish to SNS
+    /// </summary>
+    public const string SNS_PUBLISHER = "SNS";
+    /// <summary>
+    /// Specifies a publisher that will publish to EventBridge
+    /// </summary>
+    public const string EVENTBRIDGE_PUBLISHER = "EventBridge";
```
martincostello, line 22: Not that you changed them, but these constant symbol names are quite...Java-y.

src/AWS.Messaging/Configuration/PublisherMapping.cs
```diff
@@ -1,41 +1,41 @@
...
+    /// <param name="publishTargetType">The type of publisher to use</param>
+    /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
+    public PublisherMapping(Type messageType, IMessagePublisherConfiguration publisherConfiguration, string publishTargetType, string? messageTypeIdentifier = null)
+    {
+        PublishTargetType = publishTargetType;
+        MessageType = messageType;
+        MessageTypeIdentifier =
+            !string.IsNullOrEmpty(messageTypeIdentifier) ?
+            messageTypeIdentifier :
+            messageType.FullName ?? throw new InvalidMessageTypeException("Unable to retrieve the Full Name of the provided Message Type.");
```
martincostello, line 38: Would be good to add what type that was to the exception message.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -171,10 +171,8 @@ public async Task DeleteMessagesAsync(IEnumerable<MessageEnvelope> messages, Can
             }
             else
             {
-                var errorMessage = $"Attempted to delete message {message.Id} from {_configuration.SubscriberEndpoint} without an SQS receipt handle.";
-
-                _logger.LogError(errorMessage);
-                throw new MissingSQSReceiptHandleException(errorMessage);
+                _logger.LogError("Attempted to delete message {MessageId} from {SubscriberEndpoint} without an SQS receipt handle.", message.Id, _configuration.SubscriberEndpoint);
```
normj, line 174: I'm not seeing the value of duplicating the error message. We aim to use the parameterized logging methods as much as possible but since we have to format the string anyway for the exception it doesn't seem like we gain anything pushing to use the parameterized logging method.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -238,10 +236,8 @@ public async Task ExtendMessageVisibilityTimeoutAsync(IEnumerable<MessageEnvelop
             }
             else
             {
-                var errorMessage = $"Attempted to change the visibility of message {message.Id} from {_configuration.SubscriberEndpoint} without an SQS receipt handle.";
-
-                _logger.LogError(errorMessage);
-                throw new MissingSQSReceiptHandleException(errorMessage);
+                _logger.LogError("Attempted to change the visibility of message {MessageId} from {SubscriberEndpoint} without an SQS receipt handle.", message.Id, _configuration.SubscriberEndpoint);
```
normj, line 239: Same comment as above not duplicating error message.

src/AWS.Messaging/Services/HandlerInvoker.cs
```diff
@@ -65,10 +62,8 @@ public async Task<MessageProcessStatus> InvokeAsync(MessageEnvelope messageEnvel
 
             if (task == null)
             {
-                var message = $"Unexpected return type for the HandleAsync method on {subscriberMapping.HandlerType} " +
-                    $"while handling message ID {messageEnvelope.Id}. Expected {nameof(Task<MessageProcessStatus>)}";
-                _logger.LogError(message);
-                throw new InvalidMessageHandlerSignatureException(message);
+                _logger.LogError("Unexpected return type for the HandleAsync method on {HandlerType} while handling message ID {MessageEnvelopeId}. Expected {ExpectedType}", subscriberMapping.HandlerType, messageEnvelope.Id, nameof(Task<MessageProcessStatus>));
```
normj, line 65: Same comment about avoiding duplicating error message.

normj (changes requested):

src/AWS.Messaging/SQS/SQSMessagePoller.cs
slang25: Agreed that it looks weird. There is benefits of using parameterized log messages though, if we started seeing multiple of these sorts of messages in our centralized logging then we can easily filter them on the message template, once it's string interpolated and materialized to a string then we lose that and all of these logs are effectively unique when it comes to querying them.

ashovlin: Makes sense to me, admittedly when I first wrote that I was largely thinking about console logging, and not something more structured.

ashovlin (approved):
"AWS CI" is expected to fail since this is from a fork, but not too concerned this early in the project.

normj (approved):

PR #34, opened 2023-04-20
fix: extending message visibility for more than 10 entries does not fail

philasmar:
*Issue #, if available:*
DOTNET-6902

*Description of changes:*
* Added InMemory logger to be able to assert on Log messages, since most exceptions are swallowed.
* Updated the behavior to divide messages that need visibility renewal into batches of 10, which is the maximum supported by SQS.


src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -211,70 +216,96 @@ public async Task DeleteMessagesAsync(IEnumerable<MessageEnvelope> messages, Can
...
         }
 
-        var request = new ChangeMessageVisibilityBatchRequest
-        {
-            QueueUrl = _configuration.SubscriberEndpoint
-        };
+        var numberOfBatchesNeeded = (int)Math.Ceiling((decimal)messages.Count() / (decimal)SQS_MAX_MESSAGE_CHANGE_VISIBILITY);
 
-        foreach (var message in messages)
+        List<ChangeMessageVisibilityBatchRequest> requestBatches =
```
normj, line 228: This is an overly complex way of creating a collection of `ChangeMessageVisibilityBatchRequest`. Doing a fairly complex Linq statement then mod in the `for` below to figure out which `ChangeMessageVisibilityBatchRequest` to put the message in. It works but it take a while for a reader to understand what you are doing and you aren't gaining anything from simply having just the `for` loop filling up a `ChangeMessageVisibilityBatchRequest` till it reaches max then start filling in the next. Rule of thumb is prefer to simple code for readability unless making it complex gives a noticeable perf difference.

src/AWS.Messaging/Services/IMessagePoller.cs
```diff
@@ -37,5 +37,5 @@ public interface IMessagePoller
     /// </summary>
     /// <param name="messages">The messages to extend their visibility timeout.</param>
     /// <param name="token">Optional token to cancel the visibility timeout extension.</param>
-    Task ExtendMessageVisibilityTimeoutAsync(IEnumerable<MessageEnvelope> messages, CancellationToken token = default);
+    Task ExtendMessageVisibilityTimeoutAsync(List<MessageEnvelope> messages, CancellationToken token = default);
```
normj, line 40: We should be using the interface `List`. Why do we need to do this anyway? Now we have to have an extra list created.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
philasmar: Simplified code

test/AWS.Messaging.UnitTests/SQSMessagePollerTests.cs
```diff
@@ -121,6 +123,32 @@ public async Task SQSMessagePoller_ExtendMessageVisibility_Success()
             It.IsAny<CancellationToken>()));
     }
 
+    /// <summary>
+    /// Tests that calling <see cref="IMessagePoller.ExtendMessageVisibilityTimeoutAsync"/> calls
+    /// SQS's ChangeMessageVisibilityBatch with a request that has more than 10 entires.
+    /// This is expected to fail since the <see cref="ChangeMessageVisibilityBatchRequest"/> can only handle 10 entries.
```
ashovlin, line 129: nit: The _"expected to fail"_ threw me through a loop since we expect SQSMessagePoller to ultimately be successful here due to its new batching, which I think matches what is actually being asserted via the`Assert.Empty`.

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
```diff
@@ -81,6 +84,68 @@ public async Task SendAndReceive1Message()
         Assert.Equal("Test1", message.Message.MessageDescription);
     }
 
+    [Theory]
+    [InlineData(5, 1, 5)]
+    [InlineData(8, 1, 5)]
+    [InlineData(15, 1, 15)]
+    [InlineData(20, 3, 15)]
```
ashovlin, line 91: It's not obvious what scenario you had it mind for each `InlineData`. Can you add a comment to each row about why it's interesting e.g. `// Tests that visibility timeout is extended appropriately`, `// Tests that the poller waits appropriately and picks up a subsequent batch of messages`.

I want to leave some breadcrumbs so that if one of these start failing in the future the investigator knows where to start looking.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -240,37 +254,57 @@ public async Task ExtendMessageVisibilityTimeoutAsync(IEnumerable<MessageEnvelop
...
-                throw;
+                if (changeMessageVisibilityBatchTask.Exception?.InnerException is AmazonSQSException amazonEx)
+                {
+                    _logger.LogError(amazonEx, "Failed to extend the visibility of message(s) [{MessageIds}] on queue {SubscriberEndpoint}",
+                        string.Join(", ", messages.Select(x => x.Id)), _configuration.SubscriberEndpoint);
+
+                    // Rethrow the exception to fail fast for invalid configuration, permissioning, etc.
+                    if (IsSQSExceptionFatal(amazonEx))
+                    {
+                        throw amazonEx;
```
normj, line 299: Replace `throw amazonEx;` with just `throw;` so we don't reset the call stack.

normj (approved):
I'm good after my small comment and you address Alex's comments.

test/AWS.Messaging.UnitTests/SQSMessagePollerTests.cs
philasmar: You are correct. I was doing TDD and expected the test to fail before doing the fix, but I forgot to update the comment after applying the fix.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
philasmar: Doing that gives the following error: `A throw statement with no arguments is not allowed outside of a catch clause`.
Not sure I understood the reasoning behind your comment.

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
philasmar: Done

src/AWS.Messaging/SQS/SQSMessagePoller.cs
normj: I was mistaken. The diff confused me and I assumed we were in a `catch` block. You can ignore this comment.

ashovlin (approved):

PR #35, opened 2023-04-21
fix: Ensure ActiveMessageCount is updated in a thread safe manner

ashovlin:
*Issue #, if available:* N/A

*Description of changes:* @philasmar pointed out when working on PR #34 that if you call `DefaultMessageManager.ProcessMessageAsync` with a sufficient number of messages and then `await` all of the resulting tasks, `ActiveMessageCount` wouldn't arrive back at `0` as expected.

My PR #24  introduced a flaw: since the `++` and `--` operators will call the getter _then_ the setter, they'd release and reacquire the lock during the increment or decrement allowing another thread to update the value in the middle of the increment/decrement.
```
public int ActiveMessageCount 
{
        get
        {
            lock (_activeMessageCountLock)
            {
                return _activeMessageCount;
            }
        }
        set
        {
            lock (_activeMessageCountLock)
            {                
                _activeMessageCount = value;
            }
        }
    }
public Task WaitAsync(TimeSpan timeout)
{
    ActiveMessageCount++;

    // do work

    ActiveMessageCount--;
}
```

This PR moves the setter to a private method to ensure the lock is held during the entire update. I also considered `Interlocked.Increment` and `Decrement`, but think that our internal team is more used to the `lock` pattern.


slang25:
It might be worth comparing notes on this stuff, we had a similar poller implementation in JustSaying 5 and moved away from it. I did a write-up about the problems it had, some of it was due to the implementation, but in general an eager check of "how many free workers do we have" the moment one has finished leads to choppy message consumption when there's back-pressure to work through, as well as all of the complications you are facing about keeping count accurately (which is essential).

The [write-up is here](https://github.com/justeattakeaway/JustSaying/issues/422#issue-378204771).

We switched out to a mechanism that pre-fetches 1 batch lazily, with the message visibility renewal process you have it might be a better approach. I see you have the possibility to allow replacing the poller, it would be handy to us if the abstraction didn't necessitate the presence of an `ActiveMessageCount`.

philasmar (commented):
I am fine with what you are doing, but I'd prefer if we write this in a more reusable and user intuitive way. That being sad, I think it would be nice to make `UpdateActiveMessageCount` a more generic method that can update any `int` in a thread-safe way. Also it would be nice if we turn this into an extension method which would replace `ActiveMessageCount++` and `ActiveMessageCount--` by `ActiveMessageCount.UpdateValue(1)` and `ActiveMessageCount.UpdateValue(-1)` respectively.

ashovlin:
@slang25 - thanks for taking a look. I agree that we've arrived at a similar approach to what you've described in https://github.com/justeattakeaway/JustSaying/issues/422#issue-378204771, we'll look over everything you shared.

normj (approved):
I'm good with the change. Obviously we might change things as we go through @slang25 links but that is a larger order change.

ashovlin:
@slang25 - I agree with Norm's comment above, I'd rather merge this now to fix an issue in the current implementation so we can unblock work in other areas. I created https://github.com/awslabs/aws-dotnet-messaging/issues/38 to track reviewing and discussing the poller design more holistically.

@philasmar - I'd prefer not to overload `++` and `--` so that it's obvious we're doing extra work for concurrency, and am not too concerned with reusability since this is internal to the `DefaultMessageManager` and not exposed on the interface. Let me know if you still disagree.

philasmar (approved):

PR #36, opened 2023-04-21
wip: Quick prototype of reorganizing poller+manager

ashovlin:
*Description of changes:* *WIP/quick prototype top drive discussion. Tests don't pass, naming TBD*

Currently we have:
1. `SQSMessagePoller`, which is responsible for
    * Running the main polling loop
    * Calling SQS to recieve messages
    * Handing them off the the manager for handling
    * Implementing the delete and message visibility extension methods for when the manager needs to call them for a given message (or batch of messages)
2. `DefaultMessageManger`, which is responsible for
    * Managing the lifecycle of a single message (handle -> delete, and keep visibility timeout active while handling). Deletion and visibility calls back into the methods of the poller to call SQS.
    * Tracking the tasks of all messages that are in the above flow, and keeping the count so the poller knows how many to pull in its next iteration through the loop.
image

This requires both instances to have a reference to each other, which led to some uncertainty when starting the Lambda message pump. 

This draft explores refactoring to realign responsibilities:
1. `DefaultMessagePoller`, which is responsible for
    * Running the main polling loop
    * Tracking the count + tasks of all messages that are being processed
    * Managing the lifecycle of a single message (handle -> delete, and keep visibility timeout active while handling).
   
2. `SQSPoller`, which is responsible for
    * Implementing the receive, delete and message visibility extension methods 

image

For a Lambda scenario, we may able to reuse the `DefaultMessagePoller` as is and only swap out `SQSPoller` with a Lambda-specific context.


ashovlin:
Discarding, think we have it figured out in https://github.com/awslabs/aws-dotnet-messaging/pull/40

PR #39, opened 2023-04-26
docs: Add design for how the SQS poller will handle message visibility timeout for in flight messages

ashovlin:
*Issue #, if available:* DOTNET-6901

*Description of changes:* Adds a design doc for the SQS poller will manage the visibility timeout for messages that are still being handled.
* This is discussed briefly in the overall design, but here's more detail. I imagine pieces of this can be reused for the documentation later.
* A rudimentary version of this is currently implemented in `main`/`dev` (all in flight messages are extended). A code change will follow this PR with updates to the configuration options and batching behavior.

See the two open questions at the end, thanks.

### DIFF
--- a/docs/design/message-visibility-timeout-handling.md
+++ b/docs/design/message-visibility-timeout-handling.md
@@ -1,0 +1,43 @@
+# SQS Message Visibility Timeout Handling
+
+When a consumer receives a message, Amazon SQS sets a [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html). This is the period of time during which _additional_ consumers are prevented from receiving a duplicate message. The default visibility timeout is 30 seconds. The minimum is 0 seconds. The maximum is 12 hours. 
+
+When a message handler returns `MessageProcessStatus.Success`, the framework will delete the message from the queue. When a handler returns `MessageProcessStatus.Failed`, the visibility timeout will be allowed to expire and the message will become visible to consumers again (pending interactions with other SQS settings, such as the message retention period or dead-letter queue maximum receives count).
+
+The message processing framework tracks _in flight_ messages, which are those that have been received from the SQS queue but have not yet finished processing. The framework will periodically extend the visibility timeout of these messages to prevent another consumer from receiving them.
+
+This visibility timeout extension behavior can be controlled by the following settings:
+```
+builder.AddSQSPoller("<queueURL>", options => 
+{ 
+    options.VisibilityTimeout = 30; 
+    options.VisibilityTimeoutExtensionThreshold = 5;
+    options.VisibilityTimeoutExtensionHeartbeatInterval = TimeSpan.FromSeconds(1);
+});
+```
+1. `VisibilityTimeout` - This is is the length of time in seconds that the message will not be visible to other consumers once it is received, as well as the length of time that the framework will extend the visibility timeout for messages that are still processing. The default value is 30 seconds.
+    * Note that the SQS poller will always set this when it receives messages with either the configured value or _framework default_ of 30 seconds, it will not respect the value configured on the queue.
+2. `VisibilityTimeoutExtensionThreshold` - When an in flight message is within this many seconds of becoming visible again, the framework will extend its visibility timeout automatically. The new visibility timeout will be set to `VisibilityTimeout` relative to now. The default value is 5 seconds.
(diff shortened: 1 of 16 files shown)


docs/design/message-visibility-timeout-handling.md
```diff
@@ -0,0 +1,43 @@
...
+builder.AddSQSPoller("<queueURL>", options => 
+{ 
+    options.VisibilityTimeout = 30; 
+    options.VisibilityTimeoutExtensionThreshold = 5;
+    options.VisibilityTimeoutExtensionHeartbeatInterval = TimeSpan.FromSeconds(1);
+});
+```
+1. `VisibilityTimeout` - This is is the length of time in seconds that the message will not be visible to other consumers once it is received, as well as the length of time that the framework will extend the visibility timeout for messages that are still processing. The default value is 30 seconds.
+    * Note that the SQS poller will always set this when it receives messages with either the configured value or _framework default_ of 30 seconds, it will not respect the value configured on the queue.
+2. `VisibilityTimeoutExtensionThreshold` - This is how many seconds before which an in flight message would become visible again that the visibility timeout will be extended by the framework. The default value is 5 seconds.
```
philasmar, line 20: the wording of this sentence is a bit confusing.

docs/design/message-visibility-timeout-handling.md
```diff
@@ -0,0 +1,43 @@
...
+3. `VisibilityTimeoutExtensionHeartbeatInterval` - This is how frequently the framework will check in flight messages and extend the the visibility timeout of messages that are within the `VisibilityTimeoutExtensionThreshold`. The default value is 1 second.
+    * Note that is this a `TimeSpan` instead of integer seconds since it controls client behavior, and is not passed to SQS.
```
philasmar, line 22: "that is this" should be "that this is"

philasmar (approved):

docs/design/message-visibility-timeout-handling.md
```diff
@@ -0,0 +1,43 @@
+# SQS Message Visibility Timeout Handling
+
+When a consumer receives a message, Amazon SQS sets a [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html). This is the period of time during which _additional_ consumers are prevented from receiving a duplicate message. The default visibility timeout is 30 seconds. The minimum is 0 seconds. The maximum is 12 hours. 
+
+When a message handler returns `MessageProcessStatus.Success`, the framework will delete the message from the queue. When a handler returns `MessageProcessStatus.Failed`, the visibility timeout will be allowed to expire and the message will be become visible to consumers again (pending interactions with other SQS settings, such as the message retention period or dead-letter queue maximum receives count).
```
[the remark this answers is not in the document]
ashovlin: Fixed, thanks.

docs/design/message-visibility-timeout-handling.md
ashovlin: Fixed, thanks.

docs/design/message-visibility-timeout-handling.md
```diff
@@ -0,0 +1,43 @@
...
+
+The heartbeat task will check _all_ in flight messages (the number of which are controlled by `options.MaxNumberOfConcurrentMessages`) at each heartbeat interval and batch the extensions together for messages within the threshold via [ChangeMessageVisibilityBatch](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibilityBatch.html).
+
+The message visibility timeout extension feature is only available to the SQS poller configured via `AddSQSPoller`, and **not** to the forthcoming Lambda message pump. When [using Lambda with Amazon SQS](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html), Lambda will use the queue's configured visibility timeout, and [recommends setting that to at least six times the configured function timeout](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-queueconfig). 
+
+Open Questions:
+1. When a handler returns `MessageProcessStatus.Failed`, should the framework let the remaining visibility timeout run out before the message becomes visible again? Or should the framework call `ChangeMessageVisibility` with 0 seconds so the message becomes available again immediately?
```
normj, line 38: I would say now because assuming the issue is a transient issue from a down stream service the remaining visibility timeout gives those downstream services to recover. How fast users want to retry messages is controlled by `VisibilityTimeout`. If they don't care about a quick retry then set this value high and avoid MPF from making `ChangeMessageVisibilityBatch` calls.

normj: I'm just providing my opinion here. I'm good with the content being in the design and getting outside feedback.

normj (approved):

PR #41, opened 2023-05-03
chore: add more complicated integ tests

philasmar:
*Issue #, if available:*
DOTNET-6878

*Description of changes:*
* Fix bug causing messages to not be processed when 1 message in a batch throws an exception
* Added integration tests for the following use cases:
  * Mix of message types in the queue, to ensure they are routed correctly
  * Messages that aren't using the framework
  * Messages that are using the framework but don't have a configured handler
  * Handlers that fail to handle the message successfully.


src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -108,6 +108,8 @@ private async Task PollQueue(CancellationToken token)
                 MessageAttributeNames = new List<string> { "All" }
             };
 
+            var receivedMessages = new List<Message>();
```
normj, line 111: We don't need to create the `new List<Message>` since we just replace it later on the `receivedMessages = receiveMessageResponse.Messages;` statement below. Set it to null here and then check null before doing the follow up foreach below.

normj (changes requested):

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
```diff
@@ -150,6 +152,233 @@ public async Task SendAndReceiveMultipleMessages(int numberOfMessages, int visib
...
+        source.CancelAfter(60000);
+        while (!source.IsCancellationRequested)
+        {
+            if (chatMessageTempStorage.Messages.Count == numberOfMessages &&
+                foodItemTempStorage.Messages.Count == numberOfMessages &&
+                orderInfoTempStorage.Messages.Count == numberOfMessages)
+            {
+                source.Cancel();
+                break;
+            }
```
ashovlin, line 213: When working with visibility timeout I've found it useful to either wait for the queue to empty or just wait the full `CancelAfter` period - otherwise if we accidentally let the visibility timeout expire and the message gets processed again we miss seeing that we actually exceed the expected `Messsages.Count`.

But I can tweak if necessary in the visibility timeout refinement since you're ahead of me currently.

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
```diff
@@ -150,6 +152,233 @@ public async Task SendAndReceiveMultipleMessages(int numberOfMessages, int visib
...
+        Assert.Equal(numberOfMessages, chatMessageTempStorage.Messages.Count);
+        var chatMessages = chatMessageTempStorage.Messages.Where(x => x.Message.MessageDescription.Equals($"UsingFramework")).ToList();
+        foreach (var message in chatMessages)
+        {
+            Assert.NotNull(message);
+            Assert.False(string.IsNullOrEmpty(message.Id));
+            Assert.Equal("/aws/messaging", message.Source.ToString());
+            Assert.True(message.TimeStamp > publishStartTime);
+            Assert.True(message.TimeStamp < publishEndTime);
+        }
```
ashovlin, line 300: Can we assert that we're logging the non-MPF enveloped messages appropriately?

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
```diff
@@ -150,6 +152,233 @@ public async Task SendAndReceiveMultipleMessages(int numberOfMessages, int visib
...
+
+        var processStartTime = DateTime.UtcNow;
+        await pump.StartAsync(source.Token);
+
+        source.CancelAfter(60000);
+        while (!source.IsCancellationRequested) { }
+        var timeElapsed = DateTime.UtcNow - processStartTime;
+
+        var inMemoryLogger = serviceProvider.GetRequiredService<InMemoryLogger>();
+        var errorMessages = inMemoryLogger.Logs.Where(x => x.Message.Contains("Message handling failed unexpectedly for message"));
```
ashovlin, line 375: This might be a handler bug, but I don't think we want to say "unexpectedly" for an intentional `MessageProcessStatus.Failed`? 

I've been using `unexpectedly` more for places we're falling through to a plain `Exception`.

ashovlin (changes requested):

src/AWS.Messaging/SQS/SQSMessagePoller.cs
philasmar: done

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
philasmar: Done. I went with waiting for the cancellation period. I didn't want to add additional SQS calls to check how many messages are still in the queue.

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
philasmar: The message manager is responsible for this log message which gets logged in case of an exception or `MessageProcessStatus.Failed` in the handler. As discussed, I will leave this as is pending a future revisit for error handling and messaging.

normj (approved):

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
philasmar: Done

ashovlin (approved):

PR #45, opened 2023-05-10
feat: add ability to configure message source or compute it as a fallback

philasmar:
*Issue #, if available:*
DOTNET-6903

*Description of changes:*
* Added the ability for users to configure the message source in the message bus
* If a value is not configured, the source will be computed based on the compute environment the process is running in.
* Added a design doc that explains the behavior implemented in this PR.

### DIFF
--- a/docs/design/message-source-computation-design.md
+++ b/docs/design/message-source-computation-design.md
@@ -1,0 +1,61 @@
+# Message Source Computation
+
+## Background
+
+The AWS Message Processing framework adheres to the [CloudEvents specification](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md). It wraps each application message sent by a customer with a message envelope that contains all the required attributes specified by the CloudEvents spec. One of the required attributes is the **Source** attribute. The Source attribute provides context of where the message originated and primarily needs to be configurable at a message bus level by the user. If a source is not configured by the user, a value will be computed for it based on the environment in which the process is running.
+
+## How to set the message source?
+
+The **Source** attribute is stored at the `MessageConfiguration` level and will be passed to the `MessageEnvelope` irrespective of the type of message being published. When configuring the message bus, you can use the `AddMessageSource` method to set the message source.
+
+```csharp
+builder.Services.AddAWSMessageBus(builder =>
+{
+    // This is where the message source is set
+    builder.AddMessageSource(new Uri("/fancy/backend-service", UriKind.Relative))
+
+    // continue message bus configuration below ...
+    builder.AddSQSPublisher<ChatMessage>("https://sqs.us-west-2.amazonaws.com/012345678910/TestQueue");
+});
+```
+
+In case access to the `Source` attribute is needed, the attribute is accessible through the dependency injection framework by injecting `IMessageConfiguration`. The `Source` attribute is defined as follows:
+
+```csharp
+public class MessageConfiguration : IMessageConfiguration
+{
+    /// <summary>
+    /// The relative or absolute Uri to be used as a message source.
+    /// This source is added globally to any message sent through the framework.
+    /// </summary>
+    public Uri? Source { get; set; }
+}
+```
+
+## What if I don't set the message source?
+
+When a user does not specify any source attribute, the message processing framework will resolve a source depending on the compute environment that the process is executing in. The following compute environments are considered when computing the source.
+
+### AWS Lambda
+
+Lambda runtimes set several environment variables during initialization. We will look for the `AWS_LAMBDA_FUNCTION_NAME` environment variable to check if the process is running inside a Lambda function. More information on environment variables defined by AWS Lambda can be found at https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+
+**Resolved Source** - `/AWSLambda/{FUNCTION-NAME}`
+
+### Amazon ECS
+
+The Amazon ECS container agent injects an environment variable called `ECS_CONTAINER_METADATA_URI` into each container in an ECS task. To retrieve the metadata related to an ECS task we can issue a GET request to `${ECS_CONTAINER_METADATA_URI}/task`. The GET request will return a JSON response representing a dictionary. The keys in this dictionary that we are interested in are the ECS `Cluster` and `TaskARN`. More information on the task metadata endpoint in Amazon ECS can be found at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html.
+
+**Resolved Source** - `/AmazonECS/{CLUSTER}/{TASK-ARN}`
+
+### Amazon EC2
+
+The AWS SDK for .NET exposes a utility class that represents the instance metadata in EC2 which is called `EC2InstanceMetadata`. This class contains static properties that provide access to metadata from a running EC2 instance. If this class is used on a non-EC2 instance, the properties in this class will be null. We can retrieve the EC2 instance ID using the `EC2InstanceMetadata.InstanceId` property.
+
+**Resolved Source** - `/AmazonEC2/{INSTANCE-ID}`
+
+### Fallback Source Attribute
+
+If we cannot resolve the source attribute from the above compute environments then we will retrieve the DNS host name which is made available in .NET through `Dns.GetHostName()`.
+
+**Resolved Source** - `/DNSHostName/{Dns.GetHostName()}`
--- a/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
@@ -60,4 +60,18 @@ public interface IMessageBusBuilder
     /// </summary>
     /// <param name="serializationCallback">An instance of <see cref="ISerializationCallback"/>that lets users inject their own metadata to incoming and outgoing messages.</param>
     IMessageBusBuilder AddSerializationCallback(ISerializationCallback serializationCallback);
+
+    /// <summary>
+    /// Adds a global message source to the message bus.
+    /// This source will be added to the message envelope of all the messages sent through the framework.
+    /// </summary>
+    /// <param name="messageSource">The relative or absolute Uri to be used as a message source.</param>
+    IMessageBusBuilder AddMessageSource(string messageSource);
+
+    /// <summary>
+    /// Appends a suffix to the message source that is either set using <see cref="AddMessageSource(string)"/>
+    /// or computed by the framework in the absence of a user-defined source.
+    /// </summary>
+    /// <param name="suffix">The suffix to append to the message source.</param>
+    IMessageBusBuilder AddMessageSourceSuffix(string suffix);
 }
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -16,21 +16,30 @@ namespace AWS.Messaging.Serialization;
 /// </summary>
 internal class EnvelopeSerializer : IEnvelopeSerializer
 {
+    private Uri? MessageSource { get; set; }
     private const string CLOUD_EVENT_SPEC_VERSION = "1.0";
 
     private readonly IMessageConfiguration _messageConfiguration;
     private readonly IMessageSerializer _messageSerializer;
     private readonly IDateTimeHandler _dateTimeHandler;
     private readonly IMessageIdGenerator _messageIdGenerator;
+    private readonly IMessageSourceHandler _messageSourceHandler;
     private readonly ILogger<EnvelopeSerializer> _logger;
 
-    public EnvelopeSerializer(ILogger<EnvelopeSerializer> logger, IMessageConfiguration messageConfiguration, IMessageSerializer messageSerializer, IDateTimeHandler dateTimeHandler, IMessageIdGenerator messageIdGenerator)
+    public EnvelopeSerializer(
+        ILogger<EnvelopeSerializer> logger,
+        IMessageConfiguration messageConfiguration,
+        IMessageSerializer messageSerializer,
+        IDateTimeHandler dateTimeHandler,
+        IMessageIdGenerator messageIdGenerator,
+        IMessageSourceHandler messageSourceHandler)
     {
         _logger = logger;
         _messageConfiguration = messageConfiguration;
         _messageSerializer = messageSerializer;
         _dateTimeHandler = dateTimeHandler;
         _messageIdGenerator = messageIdGenerator;
+        _messageSourceHandler = messageSourceHandler;
     }
 
     /// <inheritdoc/>
@@ -38,7 +47,6 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
     {
         var messageId = await _messageIdGenerator.GenerateIdAsync();
         var timeStamp = _dateTimeHandler.GetUtcNow();
-        var source = new Uri("/aws/messaging", UriKind.Relative); // TODO: This is a dummy value. The actual value will come via the publisher configuration (pending implementation).
 
         var publisherMapping = _messageConfiguration.GetPublisherMapping(typeof(T));
         if (publisherMapping is null)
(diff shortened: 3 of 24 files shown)


docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+Lambda runtimes set several environment variables during initialization. We will look for the `AWS_LAMBDA_FUNCTION_NAME` environment variable to check if the process is running inside a Lambda function. More information on environment variables defined by AWS Lambda can be found at https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+
+**Resolved Source** - `/AWSLambdaFunction/{FUNCTION-NAME}`
+
+### Amazon ECS
+
+The Amazon ECS container agent injects an environment variable called `ECS_CONTAINER_METADATA_URI` into each container in aN ECS task. To retrieve the metadata related to an ECS task we can issue a GET request to `${ECS_CONTAINER_METADATA_URI}/task`. The GET request will return a JSON response representing a dictionary. One of the returned keys in this dictionary is the ECS `TaskARN`. More information on the task metadata endpoint in Amazon ECS can be found at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html.
+
+**Resolved Source** - `/AmazonECSTask/{TASK-ARN}`
```
normj, line 49: I'm wondering if we should add the cluster name to put a name that is more permanent then a task which can be transitory. What would you think of 

`/AmazonECS/{cluster-name}/{task-arn}`

docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+## What if I don't set the message source?
+
+When a user does not specify any source attribute, the message processing framework will resolve a source depending on the compute environment that the process is executing in. The following compute environments are considered when computing the source.
+
+### AWS Lambda
+
+Lambda runtimes set several environment variables during initialization. We will look for the `AWS_LAMBDA_FUNCTION_NAME` environment variable to check if the process is running inside a Lambda function. More information on environment variables defined by AWS Lambda can be found at https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+
+**Resolved Source** - `/AWSLambdaFunction/{FUNCTION-NAME}`
```
normj, line 43: Based on my comment about ECS where the source could have multiple I think we can have the prefix be just the service. Replace `/AWSLambdaFunction` with `/AWSLambda`

docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+The Amazon ECS container agent injects an environment variable called `ECS_CONTAINER_METADATA_URI` into each container in aN ECS task. To retrieve the metadata related to an ECS task we can issue a GET request to `${ECS_CONTAINER_METADATA_URI}/task`. The GET request will return a JSON response representing a dictionary. One of the returned keys in this dictionary is the ECS `TaskARN`. More information on the task metadata endpoint in Amazon ECS can be found at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html.
+
+**Resolved Source** - `/AmazonECSTask/{TASK-ARN}`
+
+### Amazon EC2
+
+The AWS SDK for .NET exposes a utility class that represents the instance metadata in EC2 which is called `EC2InstanceMetadata`. This class contains static properties that provide access to metadata from a running EC2 instance. If this class is used on a non-EC2 instance, the properties in this class will be null. We can retrieve the EC2 instance ID using the `EC2InstanceMetadata.InstanceId` property.
+
+**Resolved Source** - `/AmazonEC2Instance/{INSTANCE-ID}`
```
normj, line 55: Based on my comment about ECS where the source could have multiple I think we can have the prefix be just the service. Replace `/AmazonEC2Instance` with `/AmazonEC2`

src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
```diff
@@ -60,4 +60,11 @@ public interface IMessageBusBuilder
     /// </summary>
     /// <param name="serializationCallback">An instance of <see cref="ISerializationCallback"/>that lets users inject their own metadata to incoming and outgoing messages.</param>
     IMessageBusBuilder AddSerializationCallback(ISerializationCallback serializationCallback);
+
+    /// <summary>
+    /// Adds a global message source to the message bus.
+    /// This source will be added to the message envelope of all the messages sent through the framework.
+    /// </summary>
+    /// <param name="messageSource">The relative or absolute Uri to be used as a message source.</param>
+    IMessageBusBuilder AddMessageSource(Uri messageSource);
```
normj, line 69: Can we also have a `AddMessageSourceSuffix` for users that want to add additional data but would like to keep in place the auto computed value?

src/AWS.Messaging/Services/IMessageSourceHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+
+/// <summary>
+/// This interface provides the functionality to compute the message source
+/// based on the hosting environment.
+/// </summary>
+internal interface IMessageSourceHandler
+{
+    /// <summary>
+    /// Resolves a message source depending on the compute environment that it's executing in.
+    /// The following compute environments are considered, in the specified order below:
```
normj, line 17: Since this is the interface and not the implementation you might want to say this is the default search path.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -50,7 +57,7 @@ public async ValueTask<MessageEnvelope<T>> CreateEnvelopeAsync<T>(T message)
         return new MessageEnvelope<T>
         {
             Id = messageId,
-            Source = source,
+            Source = _messageConfiguration.Source ?? await _messageSourceHandler.ComputeMessageSource(),
```
normj, line 60: The `ComputeMessageSource` call is an expensive call and won't change between messages. So we should make sure we are only calling this once.

src/AWS.Messaging/Services/MessageSourceHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+        if (string.IsNullOrEmpty(messageSource))
+        {
+            messageSource = TryGetSourceFromEC2();
+        }
+        if (string.IsNullOrEmpty(messageSource))
+        {
+            messageSource = $"/DNSHostName/{_dnsManager.GetHostName()}";
+        }
+
+        _messageConfiguration.Source = new Uri(messageSource, UriKind.Relative);
```
normj, line 60: That is a weird side effect behavior here that update the config and return back the computed source. It doesn't seem like that should be the responsibility of the `IMessageSourceHandler` to handle the caching as well. That would mean custom implementations of `IMessageSourceHandler` would also have to know they need to update the `_messageConfiguration.Source` field.

src/AWS.Messaging/Services/MessageSourceHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        return _messageConfiguration.Source;
+    }
+
+    /// <summary>
+    /// Lambda runtime sets several environment variables during initialization.
+    /// This checks the AWS_LAMBDA_FUNCTION_NAME environment variable to tell if the process is running inside a Lambda function.
+    /// </summary>
+    /// <returns>Message source from AWS Lambda</returns>
+    private string TryGetSourceFromLambda()
```
normj, line 71: `Try` methods have a specific expected pattern in .NET. They should follow something like:

`bool TryXXX(<optional-key-parameter>, out <type> output)`.

If you don't want to follow that pattern that is fine but don't prefix the methods with `Try`. You can just call them `Get` methods that can return back null. In that case I would have them methods return `string?` and stop returning back empty string if not value was found.

src/AWS.Messaging/Services/MessageSourceHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// To retrieve the metadata related to an ECS task we can issue a GET request to ${ECS_CONTAINER_METADATA_URI}/task
+    /// It send back a JSON response with the TaskARN as one of the keys.
+    /// If a value is found for the TaskARN, this would indicate that the process is running in Amazon ECS.
+    /// </summary>
+    /// <returns>Message source from Amazon ECS</returns>
+    private async Task<string> TryGetSourceFromECS()
+    {
+        _logger.LogTrace("Checking if process if running in Amazon ECS...");
+
+        var ecsMetadataURI = _environmentManager.GetEnvironmentVariable("ECS_CONTAINER_METADATA_URI");
```
normj, line 94: Put the looking up the `ECS_CONTAINER_METADATA_URI` variable inside the `ECSContainerMetadataManager` type. This class doesn't need to know the details of looking up ECS data.

normj (changes requested):

docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
...
+
+### AWS Lambda
+
+Lambda runtimes set several environment variables during initialization. We will look for the `AWS_LAMBDA_FUNCTION_NAME` environment variable to check if the process is running inside a Lambda function. More information on environment variables defined by AWS Lambda can be found at https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+
+**Resolved Source** - `/AWSLambdaFunction/{FUNCTION-NAME}`
+
+### Amazon ECS
+
+The Amazon ECS container agent injects an environment variable called `ECS_CONTAINER_METADATA_URI` into each container in aN ECS task. To retrieve the metadata related to an ECS task we can issue a GET request to `${ECS_CONTAINER_METADATA_URI}/task`. The GET request will return a JSON response representing a dictionary. One of the returned keys in this dictionary is the ECS `TaskARN`. More information on the task metadata endpoint in Amazon ECS can be found at https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html.
```
ashovlin, line 47: nit: Typo in `aN`

src/AWS.Messaging/Services/DnsManager.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace AWS.Messaging.Services;
+
+/// <summary>
+/// A wrapper around <see cref="Dns"/>.
+/// </summary>
+internal class DnsManager : IDnsManager
+{
+    /// <inheritdoc/>
+    public string GetHostName() => Dns.GetHostName();
```
ashovlin, line 14: Docs say this can throw a `SocketException` - should we be more defensive?

docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
+# Message Source Computation
+
+## Background
+
+The AWS Message Processing framework adheres to the [CloudEvents specification](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md). It wraps each application message sent by a customer with a message envelope that contains all the required attributes specified by the CloudEvents spec. One of the required attributes is the **Source** attribute. The Source attribute provides context of where the message originated and primarily needs to be configurable at a message bus level by the user. If a source is not configured by the user, a value will be computed for it based on the environment in which the process is running.
+
+## How to set the message source?
+
+The **Source** attribute is stored at the `MessageConfiguration` level and will be passed to the `MessageEnvelope` irrespective of the type of message being published. An extension method is provided by the Message Processing Framework to configure the message source. When configuring the message bus, you can use the `AddMessageSource` method to set the message source.
```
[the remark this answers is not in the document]
philasmar: Done

docs/design/message-source-computation-design.md
```diff
@@ -0,0 +1,61 @@
...
+{
+    // This is where the message source is set
+    builder.AddMessageSource(new Uri("/fancy/backend-service", UriKind.Relative))
+
+    // continue message bus configuration below ...
+    builder.AddSQSPublisher<ChatMessage>("https://sqs.us-west-2.amazonaws.com/012345678910/TestQueue");
+});
+```
+
+In case access to the `Source` attribute is needed, the attribute can be found on the `MessageConfiguration` object which is accessible through the dependency injection framework by injecting `IMessageConfiguration`. The `Source` attribute is defined as follows:
```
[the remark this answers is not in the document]
philasmar: Done

docs/design/message-source-computation-design.md
philasmar: Done

docs/design/message-source-computation-design.md
philasmar: Done

docs/design/message-source-computation-design.md
philasmar: Done

src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
philasmar: Done

src/AWS.Messaging/Services/IMessageSourceHandler.cs
philasmar: Removed the implementation detail from the interface and added it on the implementation class method.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
philasmar: Done

src/AWS.Messaging/Services/MessageSourceHandler.cs
philasmar: Done

src/AWS.Messaging/Services/MessageSourceHandler.cs
philasmar: Removed the `Try` and used `Get`

src/AWS.Messaging/Services/MessageSourceHandler.cs
philasmar: Done

docs/design/message-source-computation-design.md
philasmar: Fixed

src/AWS.Messaging/Services/DnsManager.cs
philasmar: Agreed. I left this class as is because I want it to closely resemble the `Dns` class, but I added exception handling on the caller method.

src/AWS.Messaging/Services/MessageSourceHandler.cs
```diff
@@ -0,0 +1,168 @@
...
+    ///     we fallback to using <see cref="Dns.GetHostName"/></description>
+    /// </item>
+    /// </list>
+    /// After a source is computed, the message source suffix is appended if one is set.
+    /// </summary>
+    /// <returns>The computed message source.</returns>
+    public async Task<Uri> ComputeMessageSource()
+    {
+        if (_messageConfiguration.Source != null)
+            return new Uri($"{_messageConfiguration.Source}{_messageConfiguration.SourceSuffix}", UriKind.Relative);
```
normj, line 68: I think we should enforce there is a `/` in between source and source suffix. Also we should make sure there is not `//` if both source ends with a `/` and suffix starts with `/`.

src/AWS.Messaging/Services/MessageSourceHandler.cs
```diff
@@ -0,0 +1,168 @@
...
+            messageSource = GetSourceFromDnsHostName();
+        }
+        if (string.IsNullOrEmpty(messageSource))
+        {
+            messageSource = "/aws/messaging";
+        }
+
+        _logger.LogTrace("Computed message source is '{MessageSource}'", messageSource);
+
+        return new Uri($"{messageSource}{_messageConfiguration.SourceSuffix}", UriKind.Relative);
```
normj, line 91: For the utility method you create for making sure there is one and only one `/` between source and suffix use here as well.

normj (changes requested):

src/AWS.Messaging/Services/MessageSourceHandler.cs
philasmar: Done

normj (approved):

ashovlin (approved):

PR #47, opened 2023-05-12
feat: add ability to read configuration from appsettings.json

philasmar:
*Issue #, if available:*
DOTNET-6676

*Description of changes:*
* Add ability to configure the AWS Message Bus from appsettings.json
* Settings that can be configured from appsettings.json include SQS/SNS/EventBridge publishers, Message Handlers and SQS Pollers


sampleapps/SubscriberService/appsettings.json
```diff
@@ -0,0 +1,29 @@
...
+            }
+        ],
+        "SQSPollers": [
+            {
+                "QueueUrl": "https://sqs.us-west-2.amazonaws.com/012345678910/MPF",
+                "Options": {
+                    "MaxNumberOfConcurrentMessages": 10,
+                    "VisibilityTimeout": 20,
+                    "WaitTimeSeconds": 20,
+                    "VisibilityTimeoutExtensionHeartbeatInterval": "00:00:01",
```
normj, line 23: @ashovlin Looks weird to have to specify `TimeSpan` format here. I wonder how passionate you feel about having `VisibilityTimeoutExtensionHeartbeatInterval` be a `TimeSpan` instead of an `int` like the other properties.

src/AWS.Messaging/Configuration/ApplicationSettings.cs
```diff
@@ -0,0 +1,136 @@
...
+/// <summary>
+/// Represents the configuration of the Message Processing Framework.
+/// </summary>
+public class ApplicationSettings
+{
+    /// <summary>
+    /// The name of the section representing the configuration
+    /// of Message Processing Framework.
+    /// </summary>
+    public const string SectionName = "AWSMessageProcessingFramework";
```
normj, line 15: What about `AWS.Messaging` for something shorter and matches the package name?

src/AWS.Messaging/Configuration/ApplicationSettings.cs
```diff
@@ -0,0 +1,136 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Messaging.Configuration;
+
+/// <summary>
+/// Represents the configuration of the Message Processing Framework.
+/// </summary>
+public class ApplicationSettings
```
normj, line 9: I wonder if we need to make this class as part of our public API. I suspect you need it `public` for the configuration binding but maybe we should put this in an `Internal` namespace.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -62,7 +84,16 @@ private IMessageBusBuilder AddPublisher<TMessage>(IMessagePublisherConfiguration
...
     {
-        var subscriberMapping = new SubscriberMapping(typeof(THandler), typeof(TMessage), messageTypeIdentifier);
+        return AddMessageHandler(typeof(THandler), typeof(TMessage), messageTypeIdentifier);
+    }
+
+    private IMessageBusBuilder AddMessageHandler(Type handlerType, Type messageType, string? messageTypeIdentifier = null)
+    {
+        Type genericMessageHandler = typeof(IMessageHandler<>).MakeGenericType(messageType);
+        if (!handlerType.GetInterfaces().Any(x => x.Equals(genericMessageHandler)))
+            throw new Exception("ma akal");
```
normj, line 94: Need a better exception here

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -108,6 +139,87 @@ public IMessageBusBuilder AddSerializationCallback(ISerializationCallback serial
...
+        if (settings is null)
+            return this;
+
+        var callingAssembly = Assembly.GetCallingAssembly();
+
+        if (settings.SQSPublishers != null)
+        {
+            foreach (var sqsPublisher in settings.SQSPublishers)
+            {
+                var messageType = callingAssembly.GetType(sqsPublisher.MessageType);
```
normj, line 155: The referenced type might in an a different assembly then the calling assembly. Imagine having a shared models project for both the publisher and subscriber.

normj (changes requested):

src/AWS.Messaging/Configuration/ApplicationSettings.cs
philasmar: Updated

sampleapps/SubscriberService/appsettings.json
philasmar: I agree. I was not a fan of how `TimeSpan` is represented in JSON format. Waiting for @ashovlin to weigh in here.

src/AWS.Messaging/Configuration/ApplicationSettings.cs
philasmar: I had it a `internal` when I was implementing it, but I needed to make it `public` for the configuration binding to work.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: Done

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: Added a method that checks the loaded assemblies and not just the calling assembly.

src/AWS.Messaging/Configuration/ApplicationSettings.cs
normj: That was why I suggest putting in a namespace ending `Internal` like we do in the SDK and other libraries.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -108,6 +139,105 @@ public IMessageBusBuilder AddSerializationCallback(ISerializationCallback serial
...
+                }
+
+                AddSQSPoller(sqsPoller.QueueUrl, options);
+            }
+        }
+
+        return this;
+    }
+
+    private Type? GetTypeFromLoadedAssemblies(string typeValue)
```
normj, line 221: Searching all assemblies is too costly. I think we should follow the pattern of `Type.AssemblyQualifiedName` property and allow the assembly to specified in the name. The format is `<full-type-name>, <assembly-name>`. My suggestion is

* Check to see if the string has a comma. If so take the string as is and call `Type.GetType()`.
* If there is no assembly assume the type is coming from the calling assembly and find the type in there. `Assembly.GetCallingAssembly().GetType()`

normj (changes requested):

sampleapps/SubscriberService/appsettings.json
ashovlin: My thinking was:
* `VisibilityTimeoutExtensionHeartbeatInterval` isn't mapped to an SQS option/parameter at all, so we don't need to match their types (and I didn't want to use `TimeSpan` for the SQS options and then deal with runtime issues rounding it to integers)
* `VisibilityTimeoutExtensionHeartbeatInterval` controls the timing of the task that extends visibility timeouts for in flight messages, which is client-only and can be faster than 1 second. 

Capturing `VisibilityTimeoutExtensionHeartbeatInterval` as milliseconds would serialize easier, but then I worry about consistency with the options in seconds.

That said, I'm not too attached if you want to convert this to integer seconds, we can always adjust during preview.

sampleapps/PublisherAPI/Program.cs
```diff
@@ -9,12 +9,15 @@
...
 builder.Services.AddControllers();
-builder.Services.AddAWSMessageBus(builder =>
+builder.Services.AddAWSMessageBus(bus =>
 {
-    builder.AddSQSPublisher<ChatMessage>("https://sqs.us-west-2.amazonaws.com/012345678910/MPF");
-    builder.AddSNSPublisher<OrderInfo>("arn:aws:sns:us-west-2:012345678910:MPF");
-    builder.AddEventBridgePublisher<FoodItem>("arn:aws:events:us-west-2:012345678910:event-bus/default");
-    builder.ConfigureSerializationOptions(options =>
+    // Load configuration from appsettings.json
+    // bus.LoadConfigurationFromSettings(builder.Configuration);
```
ashovlin, line 15: Is this supposed to be uncommented? and replace lines 17-20?

If it's in there as a sample to the user, I'd just be more explicit: _"To load the configuration from appsettings.json instead of the code below, uncomment this and remove the following lines."_

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -108,6 +139,105 @@ public IMessageBusBuilder AddSerializationCallback(ISerializationCallback serial
...
+        if (settings is null)
+            return this;
+
+        if (settings.SQSPublishers != null)
+        {
+            foreach (var sqsPublisher in settings.SQSPublishers)
+            {
+                var messageType = GetTypeFromLoadedAssemblies(sqsPublisher.MessageType);
+                if (messageType is null)
+                    continue;
```
ashovlin, line 155: Could we log an error before continuing?

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: I am worried that we are not covering enough use cases with the proposed solution as this did not work on our unit tests. I don't know how often the fully qualified assembly name is provided, since in our examples we just specify the type with the assembly. Concerning the current implementation being too costly, how problematic is this given that this code is only ran once during startup, should the uses choose to configure the bus from appsettings.json?

sampleapps/PublisherAPI/Program.cs
philasmar: Yes this is meant to be a sample. I have updated the wording as you proposed.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -62,7 +84,16 @@ private IMessageBusBuilder AddPublisher<TMessage>(IMessagePublisherConfiguration
...
     {
-        var subscriberMapping = new SubscriberMapping(typeof(THandler), typeof(TMessage), messageTypeIdentifier);
+        return AddMessageHandler(typeof(THandler), typeof(TMessage), messageTypeIdentifier);
+    }
+
+    private IMessageBusBuilder AddMessageHandler(Type handlerType, Type messageType, string? messageTypeIdentifier = null)
+    {
+        Type genericMessageHandler = typeof(IMessageHandler<>).MakeGenericType(messageType);
+        if (!handlerType.GetInterfaces().Any(x => x.Equals(genericMessageHandler)))
+            throw new Exception("The handler type should implement 'IMessageHandler<messageType>'.");
```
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: Done

normj (approved):

Issue #46, opened 2023-05-11
Question: Is it possible to use the scoped services within a message poller ?


madhub:
For ASP.NET core REST API , we have way to inject **scoped dependencies** , its good to provide similar capability for message poller.
Something similar to [Use scoped services within a BackgroundService](https://learn.microsoft.com/en-us/dotnet/core/extensions/scoped-service). Scope can be individual message pulled from queue or a group of message pulled.

normj:
Excellent point, we should create a scope message handler invocation.

ericsampson:
was this resolved by the linked PR?

madhub:
> was this resolved by the linked PR?

Yes, resolved, closing the issue now

PR #49, opened 2023-05-26
chore: Register message handlers as scoped dependencies in the service collection

96malhar:
*Issue #, if available:*
DOTNET-6957, https://github.com/awslabs/aws-dotnet-messaging/issues/46

*Description of changes:*
This PR registers message handlers as scoped dependencies in the service collection.

### DIFF
--- a/src/AWS.Messaging/Services/HandlerInvoker.cs
+++ b/src/AWS.Messaging/Services/HandlerInvoker.cs
@@ -4,6 +4,7 @@
 using System.Collections.Concurrent;
 using System.Reflection;
 using AWS.Messaging.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
 namespace AWS.Messaging.Services;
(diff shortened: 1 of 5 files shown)


test/AWS.Messaging.UnitTests/HandlerInvokerTests.cs
```diff
@@ -103,6 +104,40 @@ public async Task HandlerInvoker_UnwrapsTargetInvocationException()
...
+        // ARRANGE
+        var serviceCollection = new ServiceCollection()
+            .AddAWSMessageBus(builder =>
+            {
+                builder.AddMessageHandler<GreetingHandler, string>("sqsQueueUrl");
+            });
+
+        var greeterMock = new Mock<IGreeter>();
+        greeterMock.Setup(g => g.Greet()).Returns("Hello!");
+        serviceCollection.AddSingleton(greeterMock.Object);
```
normj, line 124: This test isn't actually confirming that a scope was created. As far as this is concern everything could be added as transient.

IGreeter should be added as a scoped service. Then your `GreetingHandler` you should inject the `IServiceProvider` and see if multiple calls for `GetRequiredService` on `IGreeter` are the same instance but a second invoke of the handler returns back different instance of `IGreeter`.

normj (changes requested):

test/AWS.Messaging.UnitTests/HandlerInvokerTests.cs
96malhar: Done

test/AWS.Messaging.UnitTests/HandlerInvokerTests.cs
```diff
@@ -103,6 +104,48 @@ public async Task HandlerInvoker_UnwrapsTargetInvocationException()
...
+            });
+
+        serviceCollection.AddScoped<IGreeter, Greeter>();
+        serviceCollection.AddSingleton<TempStorage<string>>();
 
+        var serviceProvider = serviceCollection.BuildServiceProvider();
+        var handlerInvoker = new HandlerInvoker(serviceProvider, new NullLogger<HandlerInvoker>());
+
+        // ACT and ASSERT - Retrieve the IGreeter instance from the serviceProvider and verify that all instances are the same.
+        var greeter1 = serviceProvider.GetRequiredService<IGreeter>();
```
normj, line 129: This check should be inside the handler. You are testing the scope of the unit test not of the handler. Put this logic in the constructor of the `GreetingHandler` and then use the handlerInvoker from here.

Your handler should be exercising the ideas of a scope service but it is currently not doing anything different then a transient dependency would do.

normj (changes requested):

test/AWS.Messaging.UnitTests/HandlerInvokerTests.cs
96malhar: Moved the check to the constructor of the `GreetingHandler` :thumbsup:

normj (approved):

ashovlin (approved):

src/AWS.Messaging/Services/HandlerInvoker.cs
```diff
@@ -34,59 +35,62 @@ public HandlerInvoker(IServiceProvider serviceProvider, ILogger<HandlerInvoker>
     /// <inheritdoc/>
     public async Task<MessageProcessStatus> InvokeAsync(MessageEnvelope messageEnvelope, SubscriberMapping subscriberMapping, CancellationToken token = default)
     {
-        var handler = _serviceProvider.GetService(subscriberMapping.HandlerType);
-
-        if (handler == null)
+        using (var scope = _serviceProvider.CreateScope())
```
danielmarbach, line 38: The more future-proof version of this would be to create an [async scope](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.createasyncscope).

PR #52, opened 2023-06-01
test: Add integration tests to exercise the Lambda subscriptions

ashovlin:
*Issue #, if available:* DOTNET-6922

*Description of changes:* Adds integration tests that create actual SQS queues mapped to Lambda functions.

The technique here was lifted from https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/test/Amazon.Lambda.RuntimeSupport.Tests/Amazon.Lambda.RuntimeSupport.IntegrationTests, where we're packaging the Lambda code via an msbuild target, and then deploying it during the test setup.

The majority of this is the scaffolding, right now the tests are just testing etiher a single successful or single failing message against `ProcessLambdaEventAsync` and  `ProcessLambdaEventWithBatchResponseAsync`. But it's easier to iterate on tests now that the setup is in place, so open to feedback!

### DIFF
--- a/test/AWS.Messaging.IntegrationTests/AWS.Messaging.IntegrationTests.csproj
+++ b/test/AWS.Messaging.IntegrationTests/AWS.Messaging.IntegrationTests.csproj
@@ -4,9 +4,17 @@
     <TargetFramework>net6.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <Nullable>enable</Nullable>
+
+    <!-- Set this to true to package the Lambda functions used in the integration tests.
+    You may want to set it to false when iterating on test code but not the Lambda code. -->
+    <PackageTestLambdas>true</PackageTestLambdas>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AWSSDK.CloudWatchLogs" Version="3.7.104.72" />
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="3.7.100.128" />
+    <PackageReference Include="AWSSDK.Lambda" Version="3.7.109.6" />
+    <PackageReference Include="AWSSDK.S3" Version="3.7.104.13" />
     <PackageReference Include="AWSSDK.SecurityToken" Version="3.7.101.47" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.1" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.1" />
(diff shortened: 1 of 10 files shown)


test/AWS.Messaging.IntegrationTests/AWS.Messaging.IntegrationTests.csproj
```diff
@@ -30,4 +38,11 @@
     <ProjectReference Include="..\AWS.Messaging.Tests.Common\AWS.Messaging.Tests.Common.csproj" />
   </ItemGroup>
 
+  
+  <Target Name="PackageTestFunction" AfterTargets="AfterBuild" Condition="$(PackageTestLambdas)">
```
normj, line 42: I have actually been wanting to take trick this out of Amazon.Lamda.RuntimeSupport. It makes the dev cycle slow because every time you want to do a build even if it hasn't nothing to do with this you have to do this whole step which is slow. I would recommend migrating this flow into the `LambdaIntegrationTestFixture` test fixture so it is only done when a user wants to run the integ tests.

normj (changes requested):
Just did a brief pass but wanted to give you feedback about removing the msbuild target.

test/AWS.Messaging.Tests.Common/AWSUtilities.cs
```diff
@@ -0,0 +1,371 @@
...
+                Description = "Execution role for Lambda functions for testing the .NET message processing framework",
+                AssumeRolePolicyDocument = LambdaAssumeRolePolicy
+            };
+
+            var executionRoleArn = (await iamClient.CreateRoleAsync(createRoleRequest)).Role.Arn;
+
+            // Wait for role to propagate.
+            await Task.Delay(10000);
+
+            await iamClient.AttachRolePolicyAsync(new AttachRolePolicyRequest
```
normj, line 70: Can you add a retry around this in case the role hasn't propagated yet. 10 seconds works most of the time but sometimes it can take longer.

test/AWS.Messaging.Tests.Common/AWSUtilities.cs
```diff
@@ -0,0 +1,371 @@
...
+                {
+                    await Task.Delay(2000);
+                }
+                else
+                {
+                    throw;
+                }
+            }
+        }
+        await Task.Delay(5000);
```
normj, line 183: You can pole the function status to see if it is active yet. That will be more reliable then just a sleep. Here is some code that does the polling. https://github.com/aws/aws-extensions-for-dotnet-cli/blob/master/src/Amazon.Lambda.Tools/LambdaUtilities.cs#L877

normj (changes requested):

ashovlin:
Converted to draft until I find time to rework https://github.com/awslabs/aws-dotnet-messaging/pull/52#discussion_r1218779978, other reviewers can hold off for now.

PR #55, opened 2023-09-29
fix: Only log trace messages for the case where SQSMessagePoller fails to extend a message's visibility timeout because it was recently deleted

ashovlin:
*Issue #, if available:* DOTNET-6956

*Description of changes:* There's a relatively common race condition where these happen very close to each other:
1. A message handler task finishes processing a message, and deletes it.
2. The message timeout visibility extension task attempts to extend the visibility timeout of the same message, since the handler task is not entirely finished with it.

This results in an entry in `ChangeMessageVisibilityBatchResponse.Failed` collection for the message that was just deleted.

In flight messages are stored in a `ConcurrentDictionary`, but we don't lock the work after we've read from the dictionary and are preparing and then sending the `ChangeMessageVisibilityBatchRequest`.

Rather than introduce more synchronization between deletion and extensions, we're just going to "downgrade" these error messages to trace messages since the message has likely finished processing successfully.

### DIFF
--- a/src/AWS.Messaging/SQS/SQSMessagePoller.cs
+++ b/src/AWS.Messaging/SQS/SQSMessagePoller.cs
@@ -295,8 +295,20 @@ internal class SQSMessagePoller : IMessagePoller, ISQSMessageCommunication
 
                 foreach (var failedMessage in response.Failed)
                 {
-                    _logger.LogError("Failed to extend the visibility of message {FailedMessageId} on queue {SubscriberEndpoint}: {FailedMessage}",
-                        failedMessage.Id, _configuration.SubscriberEndpoint, failedMessage.Message);
+                    // It's possible that the task that is extending the message visibility timeout of in flight messages attempts to extend
+                    // a message whose handler task has just finished and deleted the message. Rather than adding synchronization between the two
+                    // (such as stopping handlers from deleting while the extension task is running), we "downgrade" these errors to trace messages
+                    if (failedMessage.Code.Equals("ReceiptHandleIsInvalid") &&
+                        failedMessage.Message.Equals("Message does not exist or is not available for visibility timeout change", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        _logger.LogTrace("Failed to extend the visibility of message {FailedMessageId} on queue {SubscriberEndpoint} with code {Code}, which was likely deleted: {FailedMessage}",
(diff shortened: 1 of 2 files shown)


src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -295,8 +295,20 @@ public async Task ExtendMessageVisibilityTimeoutAsync(IEnumerable<MessageEnvelop
 
                 foreach (var failedMessage in response.Failed)
                 {
-                    _logger.LogError("Failed to extend the visibility of message {FailedMessageId} on queue {SubscriberEndpoint}: {FailedMessage}",
-                        failedMessage.Id, _configuration.SubscriberEndpoint, failedMessage.Message);
+                    // It's possible that the task that is extending the message visibility timeout of in flight messages attempts to extend
+                    // a message whose handler task has just finished and deleted the message. Rather than adding synchronization between the two
+                    // (such as stopping handlers from deleting while the extension task is running), we "downgrade" these errors to trace messages
+                    if (failedMessage.Code.Equals("ReceiptHandleIsInvalid") &&
+                        failedMessage.Message.Equals("Message does not exist or is not available for visibility timeout change", StringComparison.InvariantCultureIgnoreCase))
```
philasmar, line 302: checking the error message makes this logic a bit fragile. If we do a minor update to that error message and forget to change the condition here, then we break this logic. Is the `Code` enough in this situation? or can we create a code that is specific to this situation?

ashovlin: Both the `Code` and `Message` values come from SQS, so short of taking this up the service team we have to work with what's returned for now.

For the case where the receipt handle is actually invalid, we see an exception with:
* `Code == "ReceiptHandleIsInvalid"` 
* `Message == "The input receipt handle \"<handle>\" is not a valid receipt handle."`

Generally I do agree about the maintainability concerns, but I was matching on the `message` to avoid lumping the above case (or any others that share `ReceiptHandleIsInvalid`) with the one in the code where the handle is valid, but the message has already been deleted.

I could switch to just `Code` if we're okay treating those cases as traces too.

96malhar (approved):

normj (approved):

philasmar (approved):

Issue #56, opened 2023-09-29
SQSMessagePoller.IsSQSExceptionFatal should handle non existant SQS queue exception


madhub:
Currently [SQSMessagePoller](https://github.com/awslabs/aws-dotnet-messaging/blob/d826b7934268942a5c81b19e7aeacf3ba5794d2d/src/AWS.Messaging/SQS/SQSMessagePoller.cs#L127) only checks "InvalidAddress" and "AccessDenied" decided whether to continue polling or rethrow the exception. This is not sufficient , example "QueueDoesNotExistException" exception also should stop the polling.

Better design option would be provide extensibility in `SQSMessagePollerOptions`  that can take user supplied  function to decide when top stop polling
```csharp
public class SQSMessagePollerOptions {
 
   // if not set by by the user  , configure with default set.
  private Func<AmazonSQSException,bool> _FatalErrorChecker;
   // function that takes AmazonSQSException, & returns boolean
   public Func<AmazonSQSException,bool> FatalErrorChecker
    {
      get { return _FatalErrorChecker; }   
      set { _FatalErrorChecker = value; }  
    }
}


////////////////////////////////////////////////


serviceCollection.AddAWSMessageBus(builder =>
{
    builder.AddSQSPublisher<ChatMessage>(_sqsQueueUrl);
    builder.AddSQSPoller(_sqsQueueUrl, options =>
    {
        options.FatalErrorChecker = ( sqsexp ) => 
                {
                
                     if ( sqexp.ErrorCode == "InvalidAddress" || sqexp.ErrorCode == "AccessDenied" ) {
                            return true;
                     }
                     
                     if ( sqexp.statusCode == HttpStatusCode.Unauthorized ||
                          sqexp.statusCode == HttpStatusCode.Forbidden) {
                           return true;
                      }
                      
                     if ( sqexp is QueueDoesNotExistException) {
                            return true;
                      }
                      return false;
                };  
    });
 });
```
Similar to #53

ashovlin:
`QueueDoesNotExistException` has been added to the default logic here:
https://github.com/awslabs/aws-dotnet-messaging/blob/2ec157f7099c091aaf07cb89059f574e3abcca26/src/AWS.Messaging/Configuration/SQSMessagePollerConfiguration.cs#L138

And custom failure logic is now possible as well:
```
  services.AddAWSMessageBus(builder =>
  {
[...]

PR #62, opened 2023-10-15
Added fatal checker configuration option in SQSMessagePollerOptions 

madhub:
*Issue #, if available:*
#56 

*Description of changes:*

Currently [SQSMessagePoller](https://github.com/awslabs/aws-dotnet-messaging/blob/d826b7934268942a5c81b19e7aeacf3ba5794d2d/src/AWS.Messaging/SQS/SQSMessagePoller.cs#L127) only checks "InvalidAddress" and "AccessDenied" decided whether to continue polling or rethrow the exception. This is not sufficient , example "QueueDoesNotExistException" exception also should stop the polling.

Better design option would be provide extensibility in SQSMessagePollerOptions that can take user supplied function to decide when top stop polling


src/AWS.Messaging/Configuration/SQSMessagePollerConfiguration.cs
```diff
@@ -113,4 +119,22 @@ internal MessageManagerConfiguration ToMessageManagerConfiguration()
...
         };
     }
+
+    /// <summary>
+    /// <see cref="AmazonSQSException"/> error codes that should be treated as fatal and stop the poller
+    /// </summary>
+    private static readonly HashSet<string> _fatalSQSErrorCodes = new HashSet<string>
+    {
+        "InvalidAddress",   // Returned for an invalid queue URL
+        "AccessDenied"      // Returned with insufficient IAM permissions to read from the configured queue
```
ashovlin, line 129: Feel free to add `QueueDoesNotExistException` in here. Sorry for not replying to #56 sooner, but I agree it should be in the default error list.

src/AWS.Messaging/Configuration/SQSMessagePollerConfiguration.cs
```diff
@@ -87,6 +88,11 @@ internal class SQSMessagePollerConfiguration : IMessagePollerConfiguration
     /// </remarks>
     public int WaitTimeSeconds { get; init; } = DEFAULT_WAIT_TIME_SECONDS;
 
+    /// <summary>
+    /// Function that that indicated whether SQS poller should stop polling or continue;
+    /// </summary>
+    public Func<AmazonSQSException, bool> FatalErrorChecker { get; set; } = IsSQSExceptionFatal;
```
ashovlin, line 94: Suggestion to clean up documentation and naming:
```suggestion
    /// Function that indicates whether an <see cref="AmazonSQSException"/> should stop the poller or continue.
    /// </summary>
    public Func<AmazonSQSException, bool> IsSQSExceptionFatal { get; set; } = IsSQSExceptionFatal;
```

ashovlin:
One more spot to add the new configuration, can you copy it from the `SQSMessagePollerOptions` to the `SQSMessagePollerConfiguration` here:
https://github.com/awslabs/aws-dotnet-messaging/blob/d826b7934268942a5c81b19e7aeacf3ba5794d2d/src/AWS.Messaging/Configuration/MessageBusBuilder.cs#L119-L126

It's a little awkward, but we arrived here because the queueURL is configured outside of the `Action<SQSMessagePollerOptions>` since it's the primary/required input.

ashovlin (changes requested):
One more spot to add the new configuration, can you copy it from the `SQSMessagePollerOptions` to the `SQSMessagePollerConfiguration` here:
https://github.com/awslabs/aws-dotnet-messaging/blob/d826b7934268942a5c81b19e7aeacf3ba5794d2d/src/AWS.Messaging/Configuration/MessageBusBuilder.cs#L119-L126

It's a little awkward, but we arrived here because the queueURL is configured outside of the `Action<SQSMessagePollerOptions>` since it's the primary/required input.

ashovlin (approved):

normj (approved):
Thanks for the PR!

PR #64, opened 2023-10-18
chore: Throw an exception when a message group ID is not specified while publishing to FIFO queues/topics

96malhar:
*Issue #, if available:*
DOTNET-7234

*Description of changes:*
This PR throws an exception when a message group ID is not specified while publishing to FIFO queues/topics.

### DIFF
--- a/src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
+++ b/src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
@@ -19,6 +19,8 @@ internal class SNSPublisher : IMessagePublisher, ISNSPublisher
     private readonly IMessageConfiguration _messageConfiguration;
     private readonly IEnvelopeSerializer _envelopeSerializer;
 
+    private const string FIFO_SUFFIX = ".fifo";
+
     /// <summary>
     /// Creates an instance of <see cref="SNSPublisher"/>.
     /// </summary>
@@ -83,6 +85,19 @@ internal class SNSPublisher : IMessagePublisher, ISNSPublisher
             TopicArn = topicArn,
             Message = messageBody,
         };
+
+        if (topicArn.EndsWith(FIFO_SUFFIX) && string.IsNullOrEmpty(snsOptions?.MessageGroupId))
+        {
+            var errorMessage =
+                $"You are attempting to publish to a FIFO SNS topic but the request does not include a message group ID. " +
+                $"Please use {nameof(ISNSPublisher)} from the service collection to publish to FIFO topics. " +
+                $"It exposes a {nameof(PublishAsync)} method that accepts {nameof(SNSOptions)} as a parameter. " +
+                $"A message group ID must be specified via {nameof(SNSOptions.MessageGroupId)}. " +
+                $"Additionally, {nameof(SNSOptions.MessageDeduplicationId)} must also be specified if content based de-duplication is not enabled on the topic.";
+
+            _logger.LogError(errorMessage);
+            throw new InvalidFifoPublishingRequestException(errorMessage);
+        }
 
         if (snsOptions is null)
             return request;
--- a/sampleapps/PublisherAPI/Controllers/PublisherController.cs
+++ b/sampleapps/PublisherAPI/Controllers/PublisherController.cs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using AWS.Messaging;
+using AWS.Messaging.Publishers.SNS;
+using AWS.Messaging.Publishers.SQS;
 using Microsoft.AspNetCore.Mvc;
 using PublisherAPI.Models;
 
@@ -12,10 +14,14 @@ namespace PublisherAPI.Controllers;
 public class PublisherController : ControllerBase
 {
     private readonly IMessagePublisher _messagePublisher;
+    private readonly ISQSPublisher _sqsPublisher;
+    private readonly ISNSPublisher _snsPublisher;
 
-    public PublisherController(IMessagePublisher messagePublisher)
+    public PublisherController(IMessagePublisher messagePublisher, ISQSPublisher sqsPubliser, ISNSPublisher snsPublisher)
     {
         _messagePublisher = messagePublisher;
+        _sqsPublisher = sqsPubliser;
+        _snsPublisher = snsPublisher;
     }
 
     [HttpPost("chatmessage", Name = "Chat Message")]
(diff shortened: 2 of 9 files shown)


normj (approved):

src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
```diff
@@ -72,6 +74,12 @@ public async Task PublishAsync<T>(T message, SNSOptions? snsOptions, Cancellatio
 
         _logger.LogDebug("Sending the message of type '{MessageType}' to SNS. Publisher Endpoint: {Endpoint}", typeof(T), publisherEndpoint);
         var request = CreatePublishRequest(publisherEndpoint, messageBody, snsOptions);
+
+        if (publisherEndpoint.EndsWith(FIFO_SUFFIX) && string.IsNullOrEmpty(request.MessageGroupId))
+        {
+            request.MessageGroupId = Guid.NewGuid().ToString();
```
ericsampson, line 80: I thought that for FIFO, if there isn't a natural MessageGroupId then you need to send a fixed value.
If you send a random value like this, then you've entirely defeated the intended FIFO behavior AFAIK.
Or do I just need more coffee

ericsampson: cc @normj

normj: My thinking is if a user is not sending a group id the user doesn't care about the order for that message and then they can easily use the `IMessagePublisher` interface. I wanted to make the default experience just work. When they do care about order then inject the `ISQSMessagePublisher`.  Keeping the experience with FIFO queues flexible.

My train of thought is not every message sent to a FIFO queue needs to be ordered. SQS is designed for a group of messages being ordered not the entire queue. I wouldn't want this library to use a fixed group id for all messages allowing no parallelization of message processing. Fine if the user's use case requires all messages be processed in FIFO and use a single group id and in that case they would be working with the `ISQSMessagePublisher`.

A valid question is by having this change, does it make the fact that messages are being sent in as unordered surprising and should we force every user that uses the a FIFO queue always use the `ISQSMessagePublisher`.

philasmar (approved):

src/AWS.Messaging/Publishers/MessageRoutingPublisher.cs
```diff
@@ -93,4 +96,26 @@ private IMessagePublisher GetPublisherInstance(Type publisherType)
...
     }
+
+    private void CheckFifoPublishing(PublisherMapping mapping)
+    {
+        if (!mapping.PublisherConfiguration.PublisherEndpoint.EndsWith(FIFO_SUFFIX) || mapping.PublishTargetType == PublisherTargetType.EVENTBRIDGE_PUBLISHER)
+        {
+            return;
+        }
+
+        var errorMessage = $"The generic {nameof(IMessagePublisher)} cannot be used to publish to FIFO endpoints. ";
```
normj, line 107: You should tell the user to use either the ISQS... or ISNS... and they need to set the group id. Use the error message to tell the user what they should do.

src/AWS.Messaging/Publishers/MessageRoutingPublisher.cs
```diff
@@ -93,4 +96,26 @@ private IMessagePublisher GetPublisherInstance(Type publisherType)
...
+        var fifoPublisherSampleAppUrl = "https://github.com/awslabs/aws-dotnet-messaging/blob/main/sampleapps/FIFOPublisherAPI";
+
+        if (mapping.PublishTargetType == PublisherTargetType.SQS_PUBLISHER)
+        {
+            errorMessage += $"Refer to the following sample application on how to use {nameof(ISQSPublisher)} to publish to FIFO queues: {fifoPublisherSampleAppUrl}";
```
normj, line 112: Don't have error messages refer to a sample. The code will live forever but the sample and their locations will likely change. I don't want a to consider our samples part of our breaking change ruleset.

src/AWS.Messaging/Publishers/MessageRoutingPublisher.cs
normj: I would suggest moving this check into the individual `ISQSMessagePublisher` and `ISNSMessagePublisher` so that if the uses those interfaces but still doesn't provide the group id they get the error message.

normj (changes requested):

src/AWS.Messaging/Publishers/MessageRoutingPublisher.cs
96malhar: Done

src/AWS.Messaging/Publishers/MessageRoutingPublisher.cs
96malhar: Fixed.

test/AWS.Messaging.IntegrationTests/SNSPublisherTests.cs
```diff
@@ -30,28 +32,16 @@ public SNSPublisherTests()
...
-
-        var serviceCollection = new ServiceCollection();
-        serviceCollection.AddLogging();
-        serviceCollection.AddAWSMessageBus(builder =>
-        {
-            builder.AddSNSPublisher<ChatMessage>(_snsTopicArn);
-            builder.AddMessageSource("/aws/messaging");
-        });
-        _serviceProvider = serviceCollection.BuildServiceProvider();
+        await Task.CompletedTask;
```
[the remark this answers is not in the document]
96malhar: Resolving since we've altered course.

src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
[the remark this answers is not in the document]
96malhar: Resolving since we've altered course.

normj (approved):

philasmar (approved):

src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
```diff
@@ -84,6 +86,19 @@ private PublishRequest CreatePublishRequest(string topicArn, string messageBody,
...
         };
 
+        if (topicArn.EndsWith(FIFO_SUFFIX) && string.IsNullOrEmpty(snsOptions?.MessageGroupId))
+        {
+            var errorMessage =
+                $"You are attempting to publish to a FIFO SNS topic but the request does not include a message group ID. " +
+                $"Please use {nameof(ISNSPublisher)} from the service collection to publish to FIFO topics. " +
+                $"It exposes a {nameof(PublishAsync)} method that accepts {nameof(SNSOptions)} as a parameter. " +
+                $"A message group ID must be specified via {nameof(SNSOptions.MessageGroupId)}. " +
+                $"Additionaly, {nameof(SNSOptions.MessageDeduplicationId)} must also be specified if content based de-duplication is not enabled on the topic.";
```
ashovlin, line 96: nit: typo in "Additionally" (below in SQS publisher too)

ashovlin (changes requested):

src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
96malhar: Fixed

ashovlin (approved):

src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs
96malhar: Thanks  @ericsampson for leaning in. We discussed this internally and concluded that injecting random GUIDs for the message group ID might surprise users that their messages are not being processed in order. We are now throwing an exception when a user is trying to publish a message without a message group ID. The exception message informs the user on how to specify message group IDs (and other FIFO publishing tokens) as part of the publish request.

ericsampson: thanks all! <3

PR #66, opened 2023-10-26
feat: Add support for FIFO handling of messages

96malhar:
*Issue #, if available:*
DOTNET-7235

*Description of changes:*
This PR adds support for correct FIFO handling of messages.

While processing messages from a FIFO SQS queue, this PR ensures that messages that belong to the same group are processed sequentially in FIFO order. However, messages from different message groups can still be processed concurrently based on the user provided value for `SQSMessagePollerOptions.MaxNumberOfConcurrentMessages`

### DIFF
--- a/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
+++ b/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
@@ -194,7 +194,7 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
             }
 
             var groupId = messageGroups[index];
-            var task = _messageManager.ProcessMessageGroupAsync(messageGroupMapping[groupId], token);
+            var task = _messageManager.ProcessMessageGroupAsync(messageGroupMapping[groupId], groupId, token);
             taskList.Add(task);
             index++;
         }
--- a/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
+++ b/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
@@ -22,6 +22,7 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
     private readonly IEnvelopeSerializer _envelopeSerializer;
     private readonly LambdaMessageProcessorConfiguration _configuration;
     private readonly ITelemetryFactory _telemetryFactory;
+    private readonly bool _isFifoEndpoint;
 
     private readonly SQSBatchResponse _sqsBatchResponse;
 
@@ -56,6 +57,7 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
         _telemetryFactory = telemetryFactory;
 
         _sqsBatchResponse = new SQSBatchResponse();
+        _isFifoEndpoint = _configuration.SubscriberEndpoint.EndsWith(".fifo");
     }
 
     /// <summary>
@@ -80,38 +82,24 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
                     return _sqsBatchResponse;
                 }
 
-                var taskList = new List<Task>(sqsEvent.Records.Count);
-                var index = 0;
+                var messageEnvelopeResults = new List<ConvertToEnvelopeResult>();
+
                 try
                 {
-                    while (!token.IsCancellationRequested && index < sqsEvent.Records.Count)
+                    foreach (var record in sqsEvent.Records)
                     {
-                        var numberOfMessagesToRead = _configuration.MaxNumberOfConcurrentMessages - _messageManager.ActiveMessageCount;
-
-                        // If already processing the maximum number of messages, wait for at least one to complete and then try again
-                        if (numberOfMessagesToRead <= 0)
-                        {
-                            _logger.LogTrace("The maximum number of {Max} concurrent messages is already being processed. " +
-                                "Waiting for one or more to complete for a maximum of {Timeout} seconds before attempting to poll again.",
-                                _configuration.MaxNumberOfConcurrentMessages, CONCURRENT_CAPACITY_WAIT_TIMEOUT.TotalSeconds);
-
-                            await _messageManager.WaitAsync(CONCURRENT_CAPACITY_WAIT_TIMEOUT);
-                            continue;
-                        }
-
-                        var message = ConvertToStandardSQSMessage(sqsEvent.Records[index]);
+                        var message = ConvertToStandardSQSMessage(record);
                         var messageEnvelopeResult = await _envelopeSerializer.ConvertToEnvelopeAsync(message);
+                        messageEnvelopeResults.Add(messageEnvelopeResult);
 
                         trace.AddMetadata(TelemetryKeys.MessageId, messageEnvelopeResult.Envelope.Id);
                         trace.AddMetadata(TelemetryKeys.MessageType, messageEnvelopeResult.Envelope.MessageTypeIdentifier);
-
-                        // Don't await this result, we want to process multiple messages concurrently.
-                        var task = _messageManager.ProcessMessageAsync(messageEnvelopeResult.Envelope, messageEnvelopeResult.Mapping, token);
-                        taskList.Add(task);
-                        index++;
                     }
 
-                    await Task.WhenAll(taskList);
+                    if (_isFifoEndpoint)
+                        await ProcessInFifoMode(messageEnvelopeResults, token);
+                    else
+                        await ProcessInStandardMode(messageEnvelopeResults, token);
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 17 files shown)


src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -105,37 +105,67 @@ public async Task ProcessMessageAsync(MessageEnvelope messageEnvelope, Subscribe
...
 
-        // Wait for the handler to finish processing the message
+        // Sequentially process each message within the group
+        foreach (var message in messageGroup)
+        {
+            await InvokeHandler(message.Envelope, message.Mapping, token);
+        }
+
+        // Decrement the active message count by 1 indicating that the message group is done being processed.
+        UpdateActiveMessageCount(-1);
```
normj, line 145: This seems like it should be wrapped in a `finally` block so we know for sure It is decremented.

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -105,37 +105,67 @@ public async Task ProcessMessageAsync(MessageEnvelope messageEnvelope, Subscribe
...
+        ));
 
-        // Add that metadata to the dictionary of running handlers, used to extend the visibility timeout if necessary
-        _inFlightMessageMetadata.TryAdd(messageEnvelope, metadata);
+        if (_configuration.SupportExtendingVisibilityTimeout)
+            StartMessageVisibilityExtensionTaskIfNotRunning(token);
+
+        await InvokeHandler(messageEnvelope, subscriberMapping, token);
+
+        UpdateActiveMessageCount(-1);
```
normj, line 118: This seems like it should be wrapped in a `finally` block so we know for sure It is decremented.

src/AWS.Messaging/Services/InFlightMetadata.cs
```diff
@@ -8,17 +8,11 @@ namespace AWS.Messaging.Services;
...
 internal class InFlightMetadata
 {
-    internal InFlightMetadata(Task<MessageProcessStatus> handlerTask, DateTimeOffset expectedVisibilityTimeoutExpiration)
+    internal InFlightMetadata(DateTimeOffset expectedVisibilityTimeoutExpiration)
     {
-        HandlerTask = handlerTask;
         ExpectedVisibilityTimeoutExpiration = expectedVisibilityTimeoutExpiration;
     }
 
-    /// <summary>
```
normj, line 17: Can you provide some context why this was removed?

normj (changes requested):

src/AWS.Messaging/Services/InFlightMetadata.cs
96malhar: We are currently starting this handler task and then recording the in-progress `task` as part of the `InFlightMetadata`  
https://github.com/awslabs/aws-dotnet-messaging/blob/770cf84f59054c25c3fdf49f50f1dc5e91a46ee5/src/AWS.Messaging/Services/DefaultMessageManager.cs#L109-L112

However, the status of the handler `task` is never used in the visibility timeout extension logic. Further, for FIFO queues, we want to record all messages into the `InFlightMetadata` but we don't want to start the handler invocation for all messages at once. Thus the `HandlerTask` property was excluded from the `InFlightMetadata`

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Done

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Done

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -105,37 +105,67 @@ public async Task ProcessMessageAsync(MessageEnvelope messageEnvelope, Subscribe
...
+        }
 
         if (_configuration.SupportExtendingVisibilityTimeout)
             StartMessageVisibilityExtensionTaskIfNotRunning(token);
 
-        // Wait for the handler to finish processing the message
+        // Sequentially process each message within the group
+        foreach (var message in messageGroup)
+        {
+            await InvokeHandler(message.Envelope, message.Mapping, token);
```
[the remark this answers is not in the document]
96malhar: Added this logic.

test/AWS.Messaging.UnitTests/DefaultMessageManagerTests.cs
```diff
@@ -181,9 +185,9 @@ public async Task DefaultMessageManager_ExtendsVisibilityTimeout_OnlyWhenNecessa
             mockSQSMessageCommunication.VerifyDeleteMessagesAsyncWasCalledWith(laterMessage, Times.Once());
 
             // Since the message handler takes 3 seconds, verify the the visibility was extended
-            // The 1 to 3 allows for some instability around the second boundaries
-            mockSQSMessageCommunication.VerifyExtendMessageVisibilityTimeoutAsync(new[] { earlyMessage }, Times.Between(1, 3, Moq.Range.Inclusive));
-            mockSQSMessageCommunication.VerifyExtendMessageVisibilityTimeoutAsync(new[] { laterMessage }, Times.Between(1, 3, Moq.Range.Inclusive));
+            // The 1 to 4 allows for some instability around the second boundaries
```
96malhar, line 188: Currently, we are first starting the handler task and then starting the visibility timeout extension - https://github.com/awslabs/aws-dotnet-messaging/blob/770cf84f59054c25c3fdf49f50f1dc5e91a46ee5/src/AWS.Messaging/Services/DefaultMessageManager.cs#L109-L118

With this PR, the order is now reversed and sometimes the visibility was extended 4 times in this test.

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -103,56 +103,103 @@ public Task WaitAsync(TimeSpan timeout)
...
+            }
+        }
+        finally
+        {
+            // Decrement the active message count by 1 indicating that the message group is done being processed.
+            UpdateActiveMessageCount(-1);
+        }
+    }
+
+    private void acknowledgeMessage(MessageEnvelope messageEnvelope, CancellationToken token)
```
normj, line 152: Follow .NET convention and start method name with capital "A".

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -103,56 +103,103 @@ public Task WaitAsync(TimeSpan timeout)
...
+            }
+
+            // Sequentially process each message within the group
+            foreach (var message in messageGroup)
+            {
+                var isSuccessful = await InvokeHandler(message.Envelope, message.Mapping, token);
+                if (!isSuccessful)
+                {
+                    // If the handler invocation fails for any message, skip processing subsequent messages in the group.
+                    _logger.LogError("Handler invocation failed for a message belonging to message group '{groupdId}'. Skipping processing of subsequent messages from the same group.", groupId);
```
normj, line 140: Nice to have, could be helpful for debugging to log how many messages didn't get processed. Also include the message id of the message that failed.

test/AWS.Messaging.IntegrationTests/SubscriberTests.cs
```diff
@@ -356,7 +357,7 @@ public async Task MessagesWithFailedHandlers(int numberOfMessages)
         var timeElapsed = DateTime.UtcNow - processStartTime;
 
         var inMemoryLogger = serviceProvider.GetRequiredService<InMemoryLogger>();
-        var errorMessages = inMemoryLogger.Logs.Where(x => x.Message.Contains("Message handling failed unexpectedly for message"));
+        var errorMessages = inMemoryLogger.Logs.Where(x => x.Message.Contains("Message handling completed unsuccessfully for message"));
```
96malhar, line 360: This test uses the [ChatMessageHandler_Failed](https://github.com/awslabs/aws-dotnet-messaging/blob/770cf84f59054c25c3fdf49f50f1dc5e91a46ee5/test/AWS.Messaging.IntegrationTests/Handlers/ChatMessageHandler.cs#L46-L60) handler and asserts on the expected log messages when the handler invocation is completed but with a failed status. 

However, we were not adding the `TempStorage` into the service collection because of which an instance of the handler could never be constructed. This led to a premature failure without the handler ever being invoked.

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -103,56 +103,103 @@ public Task WaitAsync(TimeSpan timeout)
...
+            await handlerTask;
         }
         catch (AWSMessagingException)
         {
             // Swallow exceptions thrown by the framework, and rely on the thrower to log
         }
         catch (Exception ex)
         {
-            _logger.LogError(ex, "An unknown exception occurred while processing message ID {SubscriberEndpoint}", messageEnvelope.Id);
+            _logger.LogError(ex, "An unknown exception occurred while processing message ID {messageId}", messageEnvelope.Id);
```
normj, line 177: Error message suggestion

```
_logger.LogError(ex, "An exception has been thrown from handler {handlerType} processing message {messageId}", subscriberMapping.HandlerType.Name, messageEnvelope.Id);
```

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -103,56 +103,103 @@ public Task WaitAsync(TimeSpan timeout)
...
             {
                 _logger.LogError("Message handling completed unsuccessfully for message ID {MessageId}", messageEnvelope.Id);
                 await _sqsMessageCommunication.ReportMessageFailureAsync(messageEnvelope);
             }
         }
-        else if (metadata.HandlerTask.IsFaulted)
+        else if (handlerTask.IsFaulted)
         {
-            _logger.LogError(metadata.HandlerTask.Exception, "Message handling failed unexpectedly for message ID {MessageId}", messageEnvelope.Id);
+            _logger.LogError(handlerTask.Exception, "Message handling failed unexpectedly for message ID {MessageId}", messageEnvelope.Id);
```
normj, line 198: In this error message and the one above for an unsuccessful processing can we include the handler type name in error message like I suggested above.

normj (changes requested):

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Added this.

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Done

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Done

src/AWS.Messaging/Services/DefaultMessageManager.cs
96malhar: Done

normj (approved):

src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
```diff
@@ -138,6 +126,82 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
...
+
+            var task = _messageManager.ProcessMessageAsync(messageEnvelopeResults[index].Envelope, messageEnvelopeResults[index].Mapping, token);
+            taskList.Add(task);
+            index++;
+        }
+
+        await Task.WhenAll(taskList);
+    }
+
+    private async Task ProcessInFifoMode(List<ConvertToEnvelopeResult> messageEnvelopResults, CancellationToken token)
```
ashovlin, line 157: nit: Type in "Envelope"

src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
```diff
@@ -138,6 +126,82 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
...
+        var messageGroupMapping = new Dictionary<string, List<ConvertToEnvelopeResult>>();
+
+        foreach (var messageEnvelopResult in messageEnvelopResults)
+        {
+            var groupId = messageEnvelopResult.Envelope.SQSMetadata?.MessageGroupId;
+
+            if (string.IsNullOrEmpty(groupId))
+            {
+                // This should never happen. But if it does, its an issue with the framework. This is not a customer induced error.
+                throw new InvalidOperationException($"This SQS message cannot be processed in FIFO mode because it does not have a valid message group ID");
```
ashovlin, line 168: Should we `LogError` here? Or will something higher up do so?

Ditto for non-Lambda below.

src/AWS.Messaging/SQS/SQSMessagePoller.cs
```diff
@@ -138,25 +140,64 @@ private async Task PollQueue(CancellationToken token)
...
+                ProcessInStandardMode(messageEnvelopResults, token);
+        }
+    }
+
+    private void ProcessInStandardMode(List<ConvertToEnvelopeResult> messageEnvelopResults, CancellationToken token)
+    {
+        foreach (var result in messageEnvelopResults)
+        {
+            // Don't await this result, we want to process multiple messages concurrently
+            _ = _messageManager.ProcessMessageAsync(result.Envelope, result.Mapping, token);
```
ashovlin, line 173: The `ProcessMessageAsync` call was previously in a `try` with catches for `AWSMessagingException` and `Exception`. 

But now the `ProcessInStandardMode` invocation is after that. Does the `AWSMessagingException` still get swallowed appropriately?

src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -2,8 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Collections.Concurrent;
-using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
```
ashovlin, line 5: Minor, but why is this here? Is it unused?

ashovlin (commented):
Could you also tidy up the internal design doc and move it to the https://github.com/awslabs/aws-dotnet-messaging/tree/main/docs/design folder?

We did explore multiple approaches here, so think it'd be good to open source our thinking.

96malhar:
> Could you also tidy up the internal design doc and move it to the https://github.com/awslabs/aws-dotnet-messaging/tree/main/docs/design folder?
> 
> We did explore multiple approaches here, so think it'd be good to open source our thinking.

I'd prefer to do this separately and I've created DOTNET-7269 to track this work.

ashovlin (approved):

PR #67, opened 2023-10-27
chore: add SourceLink to project files

philasmar:
*Issue #, if available:*
DOTNET-6926

*Description of changes:*
Added the properties required for SourceLink as well as the proper package reference. Once the projects are packed using `dotnet pack`, a symbols file will be generated alongside the NuGet package. In our pipelines, when we do a `nuget push`, if there is a symbols package in the same directory as the NuGet package, then the symbols package will automatically be uploaded.

### DIFF
--- a/src/AWS.Messaging.Lambda/AWS.Messaging.Lambda.csproj
+++ b/src/AWS.Messaging.Lambda/AWS.Messaging.Lambda.csproj
@@ -17,11 +17,16 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\public.snk</AssemblyOriginatorKeyFile>
     <Version>0.1.0-beta</Version>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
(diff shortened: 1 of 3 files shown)


normj (commented):
For the snupkg files do we need to update publishing code in the pipeline to publish snupgk files?

Also I'm assuming some part of the build process with the source link project captures the current commit id. Are we always running the `dotnet build` / `dotnet pack` from the context of the Git repository?

philasmar:
> For the snupkg files do we need to update publishing code in the pipeline to publish snupgk files?
> 
> Also I'm assuming some part of the build process with the source link project captures the current commit id. Are we always running the `dotnet build` / `dotnet pack` from the context of the Git repository?

We shouldn't need to update our build/publish code since a nuget push will detect the snupkg and push it up with the regular package.
From what I;ve seen, when we do a nuget pack, we need to be in context of the .git folder for it to work.

normj (approved):

ashovlin (approved):

PR #68, opened 2023-11-01
feat: Add OpenTelemetry Instrumentation Provider

ashovlin:
*Issue #, if available:* DOTNET-7031

*Description of changes:*
* Adds `AWS.Messaging.Telemetry.OpenTelemetry`, which instruments the AWS Message Processing Framework for .NET for OpenTelemetry
    * Based heavily on @normj 's prototype in https://github.com/awslabs/aws-dotnet-messaging/tree/normj/telemetry
* Outside of OTel, the framework will now serialize `MessageEnvelope.Metadata` as top-level properties in the JSON envelope when sending messages. We're using this to telemetry, but us and customers can use for other CloudEvents extensions and/or custom attributes.

### DIFF
--- a/sampleapps/PublisherAPI/PublisherAPI.csproj
+++ b/sampleapps/PublisherAPI/PublisherAPI.csproj
@@ -7,10 +7,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.6.0" />
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.6.0" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\src\AWS.Messaging.Telemetry.OpenTelemetry\AWS.Messaging.Telemetry.OpenTelemetry.csproj" />
     <ProjectReference Include="..\..\src\AWS.Messaging\AWS.Messaging.csproj" />
   </ItemGroup>
 
--- a/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
+++ b/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics;
-using System.Text.Json;
 using OpenTelemetry;
 using OpenTelemetry.Context.Propagation;
 
@@ -13,7 +12,7 @@ namespace AWS.Messaging.Telemetry.OpenTelemetry
     /// </summary>
     public class OpenTelemetryProvider : ITelemetryProvider
     {
-        private static readonly ActivitySource _activitySource = new ActivitySource(TelemetryKeys.Source, TelemetryKeys.AWSMessagingAssemblyVersion);
+        private static readonly ActivitySource _activitySource = new ActivitySource(Constants.SourceName, TelemetryKeys.AWSMessagingAssemblyVersion);
 
         /// <inheritdoc/>
         public ITelemetryTrace Trace(string traceName)
@@ -74,16 +73,9 @@ namespace AWS.Messaging.Telemetry.OpenTelemetry
         /// <returns>Context value</returns>
         private IEnumerable<string> ExtractTraceContextFromEnvelope(MessageEnvelope envelope, string key)
         {
-            if (envelope.Metadata.TryGetValue(key, out var value))
-            {               
-                if (value is JsonElement jsonValue)
-                {
-                    return new string[] { jsonValue.ToString() };
-                }
-                else if (value is string stringValue)
-                {
-                    return new string[] { stringValue };
-                }
+            if (envelope.Metadata.TryGetValue(key, out var jsonElement))
+            {
+                return new string[] { jsonElement.ToString() };
             }
 
             return Enumerable.Empty<string>();
--- a/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
+++ b/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
@@ -5,80 +5,79 @@ using System.Diagnostics;
 using OpenTelemetry;
 using OpenTelemetry.Context.Propagation;
 
-namespace AWS.Messaging.Telemetry.OpenTelemetry
+namespace AWS.Messaging.Telemetry.OpenTelemetry;
+
+/// <summary>
+/// Creates OpenTelemetry traces
+/// </summary>
+public class OpenTelemetryProvider : ITelemetryProvider
 {
-    /// <summary>
-    /// Creates OpenTelemetry traces
-    /// </summary>
-    public class OpenTelemetryProvider : ITelemetryProvider
+    private static readonly ActivitySource _activitySource = new ActivitySource(Constants.SourceName, TelemetryKeys.AWSMessagingAssemblyVersion);
+
+    /// <inheritdoc/>
+    public ITelemetryTrace Trace(string traceName)
     {
-        private static readonly ActivitySource _activitySource = new ActivitySource(Constants.SourceName, TelemetryKeys.AWSMessagingAssemblyVersion);
-
-        /// <inheritdoc/>
-        public ITelemetryTrace Trace(string traceName)
+        var activity = _activitySource.StartActivity(traceName, ActivityKind.Producer);
+        if (activity != null)
         {
-            var activity = _activitySource.StartActivity(traceName, ActivityKind.Producer);
-            if (activity != null)
-            {
-                return new OpenTelemetryTrace(activity);
-            }
-
-            // If we initially failed to create an activity, attempt to force creation with 
-            // a link to the current activity, see https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities
-            var parentActivity = Activity.Current;
-            Activity.Current = null;
-            ActivityLink[]? links = null;
-            if (parentActivity != null)
-            {
-                links = new[] { new ActivityLink(parentActivity.Context) };
-            }
-
-            activity = _activitySource.StartActivity(traceName, ActivityKind.Producer, parentContext: default, links: links);
-            
-            return new OpenTelemetryTrace(activity, parentActivity);
+            return new OpenTelemetryTrace(activity);
         }
 
-        /// <inheritdoc/>
-        public ITelemetryTrace Trace(string traceName, MessageEnvelope envelope)
+        // If we initially failed to create an activity, attempt to force creation with 
+        // a link to the current activity, see https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities
+        var parentActivity = Activity.Current;
+        Activity.Current = null;
+        ActivityLink[]? links = null;
+        if (parentActivity != null)
         {
-            var propogatedContext = Propagators.DefaultTextMapPropagator.Extract(default, envelope, ExtractTraceContextFromEnvelope);
-            Baggage.Current = propogatedContext.Baggage;
-
-            var activity = _activitySource.StartActivity(traceName, ActivityKind.Consumer, parentContext: propogatedContext.ActivityContext);
-            if (activity != null)
-            {
-                return new OpenTelemetryTrace(activity);
-            }
-
-            // If we initially failed to create an activity, attempt to force creation with 
-            // a link to the current activity, see https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities
-            var parentActivity = Activity.Current;
-            Activity.Current = null;
-            ActivityLink[]? links = null;
-            if (parentActivity != null)
-            {
-                links = new[] { new ActivityLink(parentActivity.Context) };
-            }
-
-            activity = _activitySource.StartActivity(traceName, ActivityKind.Consumer, parentContext: propogatedContext.ActivityContext, links: links);
-
-            return new OpenTelemetryTrace(activity, parentActivity);
+            links = new[] { new ActivityLink(parentActivity.Context) };
         }
 
-        /// <summary>
-        /// Extracts propogation context from a <see cref="MessageEnvelope"/>, meant to be used with <see cref="TextMapPropagator"/>
-        /// </summary>
-        /// <param name="envelope">Inbound message envelope</param>
-        /// <param name="key">Context key</param>
-        /// <returns>Context value</returns>
-        private IEnumerable<string> ExtractTraceContextFromEnvelope(MessageEnvelope envelope, string key)
-        {
-            if (envelope.Metadata.TryGetValue(key, out var jsonElement))
-            {
-                return new string[] { jsonElement.ToString() };
-            }
+        activity = _activitySource.StartActivity(traceName, ActivityKind.Producer, parentContext: default, links: links);
+        
+        return new OpenTelemetryTrace(activity, parentActivity);
+    }
 
-            return Enumerable.Empty<string>();
+    /// <inheritdoc/>
+    public ITelemetryTrace Trace(string traceName, MessageEnvelope envelope)
+    {
+        var propogatedContext = Propagators.DefaultTextMapPropagator.Extract(default, envelope, ExtractTraceContextFromEnvelope);
+        Baggage.Current = propogatedContext.Baggage;
+
+        var activity = _activitySource.StartActivity(traceName, ActivityKind.Consumer, parentContext: propogatedContext.ActivityContext);
+        if (activity != null)
+        {
+            return new OpenTelemetryTrace(activity);
         }
+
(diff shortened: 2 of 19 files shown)


sampleapps/PublisherAPI/PublisherAPI.csproj
```diff
@@ -7,10 +7,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.6.0" />
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.6.0" />
```
ashovlin, line 11: The `OpenTelemetry.Extensions.Hosting` is necessary to use our new `AddAWSMessagingInstrumentation` extension method. Should it just be a dependency of our OTel package?

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// Extracts propogation context from a <see cref="MessageEnvelope"/>, meant to be used with <see cref="TextMapPropagator"/>
+        /// </summary>
+        /// <param name="envelope">Inbound message envelope</param>
+        /// <param name="key">Context key</param>
+        /// <returns>Context value</returns>
+        private IEnumerable<string> ExtractTraceContextFromEnvelope(MessageEnvelope envelope, string key)
+        {
+            if (envelope.Metadata.TryGetValue(key, out var value))
+            {               
+                if (value is JsonElement jsonValue)
```
ashovlin, line 79: Norm's prototype introduced [DictionaryStringObjectConverter](https://github.com/awslabs/aws-dotnet-messaging/blob/normj/telemetry/src/AWS.Messaging/Serialization/DictionaryStringObjectConverter.cs), which was deserializing these back to strings. 

I didn't implement that yet since we'd have to cover a lot of primitive types, so for now customers and us will have to use `JsonElement` and know what type corresponds to which key. Let me know if you'd rather use that approach though.

This also means the `else if` is dead code, had left it in from iterating on tests.

src/AWS.Messaging/MessageEnvelope.cs
```diff
@@ -45,7 +45,13 @@ public abstract class MessageEnvelope
 
     /// <summary>
     /// This stores different metadata that is not modeled as a top-level property in MessageEnvelope class.
+    /// These entries will also be serialized as top-level properties when sending the message, which
+    /// can be used for CloudEvents Extension Attributes.
```
ashovlin, line 49: Curious for your thoughts on this. If we did not want to give customers the ability to do this, we can move telemetry (and any other future use cases) to a `internal PrivateMetadata` member for stuff we do want to serialize, and then `Metadata` will only be used for deserialization like it was previously.

test/AWS.Messaging.UnitTests/SerializationTests/EnvelopeSerializerTests.cs
```diff
@@ -343,7 +343,7 @@ public async Task SerializationCallbacks_AreCorrectlyInvoked()
         var serializedMessage = await envelopeSerializer.SerializeAsync(messageEnvelope);
 
         // ASSERT - Check expected base 64 encoded string
-        var expectedserializedMessage = "eyJpZCI6IjEyMyIsInNvdXJjZSI6Ii9hd3MvbWVzc2FnaW5nIiwic3BlY3ZlcnNpb24iOiIxLjAiLCJ0eXBlIjoiYWRkcmVzc0luZm8iLCJ0aW1lIjoiMjAwMC0xMi0wNVQxMDozMDo1NSswMDowMCIsImRhdGEiOiJ7XHUwMDIyVW5pdFx1MDAyMjoxMjMsXHUwMDIyU3RyZWV0XHUwMDIyOlx1MDAyMlByaW5jZSBTdFx1MDAyMixcdTAwMjJaaXBDb2RlXHUwMDIyOlx1MDAyMjAwMDAxXHUwMDIyfSJ9";
+        var expectedserializedMessage = "eyJpZCI6IjEyMyIsInNvdXJjZSI6Ii9hd3MvbWVzc2FnaW5nIiwic3BlY3ZlcnNpb24iOiIxLjAiLCJ0eXBlIjoiYWRkcmVzc0luZm8iLCJ0aW1lIjoiMjAwMC0xMi0wNVQxMDozMDo1NSswMDowMCIsImRhdGEiOiJ7XHUwMDIyVW5pdFx1MDAyMjoxMjMsXHUwMDIyU3RyZWV0XHUwMDIyOlx1MDAyMlByaW5jZSBTdFx1MDAyMixcdTAwMjJaaXBDb2RlXHUwMDIyOlx1MDAyMjAwMDAxXHUwMDIyfSIsIklzLURlbGl2ZXJlZCI6ZmFsc2V9";
```
ashovlin, line 346: This is due to the change to now serialize `Metadata` in the outgoing message.

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
normj: I would prefer we do some conversion. It seems confusing to users that in the publish they put in a string or number and then on the subscriber they type has changed. And it feels like we are leaking our serialization logic potentially into user's business code.

src/AWS.Messaging.Telemetry.OpenTelemetry/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Configuration
+
+### 1. Install Packages
+Add a reference to [`AWS.Messaging.Telemetry.OpenTelemetry`](https://www.nuget.org/packages/AWS.Messaging.Telemetry.OpenTelemetry) and [`OpenTelemetry.Extensions.Hosting`](https://www.nuget.org/packages/OpenTelemetry.Extensions.Hosting).
+
+You may also add a reference to one or more [exporters](https://opentelemetry.io/docs/instrumentation/net/exporters/) to visualize your telemetry data.
+
+```shell
+dotnet add package --prerelease AWS.Messaging.Telemetry.OpenTelemetry
+dotnet add package OpenTelemetry.Extensions.Hosting
```
normj, line 18: I'm trying to understand if installing `OpenTelemetry.Extensions.Hosting` is just what you have to do to use OTel. I want to make it clear when we are explaining how to use OTel with .NET and when to do something special for us.

src/AWS.Messaging/MessageEnvelope.cs
normj: I do want to give customers the ability to add their own metadata. Maybe though we only need support `string` metadata though. But that gets in the way of the `JsonExtensionData`. Maybe those concerns of application level metadata should be separated from extra deserialized data place holder..

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -95,6 +95,17 @@ public async ValueTask<string> SerializeAsync<T>(MessageEnvelope<T> envelope)
                 ["data"] = _messageSerializer.Serialize(message)
             };
 
+            // Write any Metadata as top-level keys
+            // This may be useful for any extensions defined in
+            // https://github.com/cloudevents/spec/tree/main/cloudevents/extensions
+            foreach (var key in envelope.Metadata.Keys)
```
normj, line 101: Is that your read for Cloud Events that extensions always put their data at the root of the JSON document?

normj (changes requested):

sampleapps/PublisherAPI/PublisherAPI.csproj
ashovlin: `OpenTelemetry.Extensions.Hosting` enables `IServiceCollection.AddOpenTelemetry`, which is necessary to use our `AddAWSMessagingInstrumentation` extension that registers our `OpenTelemetryProivder` singleton and adds our source name.

But it is not strictly required, since you can also enable OpenTelemetry via `CreateTracerProviderBuilder` like we do in the tests. This is brought in via `OpenTelemetry`, which is a transitive dependency via our new `AWS.Messaging.Telemetry.OpenTelemetry`.
```
using (var tracerProvider = Sdk.CreateTracerProviderBuilder()
           .AddSource(TelemetryKeys.Source)
           .ConfigureResource(resource => resource.AddService("unittest"))
           .AddInMemoryExporter(activities).Build())
{
    await _publisher.PublishAsync(new ChatMessage { MessageDescription = "Test Description" });
}
```

----

So my thinking was that because the user isn't required to configure OTel via DI, we didn't want to require the dependency ourselves.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
ashovlin: Yeah, from https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/primer.md#json-extensions
> CloudEvent extension attributes are serialized as siblings to the specification defined attributes - meaning, at the top-level of the JSON object. The authors of the specification spent a long time considering all options and decided that this was the best choice.

src/AWS.Messaging.Telemetry.OpenTelemetry/README.md
ashovlin: See reply here: https://github.com/awslabs/aws-dotnet-messaging/pull/68#discussion_r1388129171

Added another line to the README explaining why the example is including it, though not absolutely necessary.

src/AWS.Messaging/Telemetry/TelemetryKeys.cs
```diff
@@ -1,11 +1,27 @@
...
-internal static class TelemetryKeys
+/// <summary>
+/// Constants related to telemetry
+/// </summary>
+public static class TelemetryKeys
 {
+    /// <summary>
+    /// OpenTelemetry activity source name
+    /// </summary>
+    public const string Source = "AWS.Messaging";
```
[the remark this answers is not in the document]
ashovlin: Moved.

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
ashovlin: I attempted to apply the converter like the prototype, but realized via https://github.com/dotnet/runtime/issues/32903 that a custom `JsonConverter` cannot deserialize the dictionary attributed with `JsonExtensionData` by design.

The prototype was working because it was doing something like this, so that `MessageEnvelope.Metadata` was not being treated as overflow via `JsonExtensionData`
```
{ 
    "metadata": {
         "traceparent": "..."
    } 
}
```

But we want the OTel info (and possibly other extensions, set by us internally or users) in the top of the CloudEvents envelope. 

We had an internal discussion on 11/9/23 debating:
1. Leave deserialization as-is (both us and users will have to work with `JsonElement` values) when handling `Metadata`. We’ll route MPF’s OTel info through this path.
2. Move the OTel properties to real properties in our `MessageEnvelope<T>`  
   * Simplifies this PR since we're not working with overflow, but not user-provided extensions. 
   * We also want to avoid any OTel code in the core `AWS.Messaging` package if possible.
3. We could use `JsonExtensionData` to initially de-serialize overflow into something like `internal RawMetadata<string, JsonElement>` and then do our own copy+casting into a separate `public Metadata<string, object>`.

We agreed to start with approach number 1 for now.

src/AWS.Messaging/MessageEnvelope.cs
ashovlin: See https://github.com/awslabs/aws-dotnet-messaging/pull/68#discussion_r1389667763

Customers can add to the envelope-level metadata via `ISerializationCallback`, but for now will have to work with `JsonElement` values.

sampleapps/PublisherAPI/PublisherAPI.csproj
normj: Lets start without it. Easier to add it then remove it.

normj (approved):

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
```diff
@@ -0,0 +1,83 @@
...
+            links = new[] { new ActivityLink(parentActivity.Context) };
+        }
+
+        activity = _activitySource.StartActivity(traceName, ActivityKind.Consumer, parentContext: propogatedContext.ActivityContext, links: links);
+
+        return new OpenTelemetryTrace(activity, parentActivity);
+    }
+
+    /// <summary>
+    /// Extracts propogation context from a <see cref="MessageEnvelope"/>, meant to be used with <see cref="TextMapPropagator"/>
```
philasmar, line 69: typo in "propogation" should be "propagation"

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryTrace.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <inheritdoc/>
+    public void RecordTelemetryContext(MessageEnvelope envelope)
+    {
+        ActivityContext contextToInject = default;
+        if (_activity != null)
+        {
+            contextToInject = _activity.Context;
+        }
+        // Even if an "AWS.Messaging" activity was not created, we still
+        // propogate the the current activity (if it exists) through the message envelope
```
philasmar, line 60: you have 2 "the" in a row.

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryTrace.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// </summary>
+    /// <param name="envelope">Outbound message envelope</param>
+    /// <param name="key">Context key</param>
+    /// <param name="value">Context value</param>
+    private void InjectTraceContextIntoEnvelope(MessageEnvelope envelope, string key, string value)
+    {
+        envelope.Metadata[key] = JsonSerializer.SerializeToElement(value);
+    }
+
+    private bool _disposed;
```
philasmar, line 80: nit: add a new line between the property and the summary below

test/AWS.Messaging.UnitTests/SerializationTests/EnvelopeSerializerTests.cs
```diff
@@ -362,7 +362,7 @@ public async Task SerializationCallbacks_AreCorrectlyInvoked()
         Assert.Equal("1.0", envelope.Version);
         Assert.Equal("/aws/messaging", envelope.Source?.ToString());
         Assert.Equal("addressInfo", envelope.MessageTypeIdentifier);
-        Assert.Equal(true, envelope.Metadata["Is-Delivered"]);
+        Assert.Equal(true, envelope.Metadata["Is-Delivered"].GetBoolean());
```
philasmar, line 365: could you please resolve the warnings?

ashovlin: I was taking a small moral stance here, since we weren't really asserting that _"some condition is true"_, rather more like _"this piece of data is equal to the value `true`"_. 

But updated, not a big deal.

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryTrace.cs
ashovlin: Fixed in https://github.com/awslabs/aws-dotnet-messaging/pull/68/commits/759236875251f757e2a10fb90851581e84ce8642

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryTrace.cs
ashovlin: Fixed in https://github.com/awslabs/aws-dotnet-messaging/pull/68/commits/759236875251f757e2a10fb90851581e84ce8642

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryTrace.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Diagnostics;
+using System.Text.Json;
+using OpenTelemetry;
+using OpenTelemetry.Context.Propagation;
+using OpenTelemetry.Trace;
+
+namespace AWS.Messaging.Telemetry.OpenTelemetry;
+
+/// <summary>
+/// An OpenTelemetry trace (wrapper around a <see cref="Activity"/>
```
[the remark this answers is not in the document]
ashovlin: Fixed in https://github.com/awslabs/aws-dotnet-messaging/pull/68/commits/759236875251f757e2a10fb90851581e84ce8642

src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
ashovlin: Fixed in https://github.com/awslabs/aws-dotnet-messaging/pull/68/commits/759236875251f757e2a10fb90851581e84ce8642

philasmar (approved):

PR #71, opened 2023-11-14
test: Add integration tests to exercise the Lambda subscriptions 

ashovlin:
*Issue #, if available:* DOTNET-6922

*Description of changes:* Adds integration tests that create actual SQS queues mapped to Lambda functions.

Addressed feedback on https://github.com/awslabs/aws-dotnet-messaging/pull/52 at the time it was closed (due to us stopping and then restarting this ticket now). I couldn't re-open since I rebased the branch on `dev`.

The majority of this is the scaffolding, right now the tests are just testing either a single successful or single failing message against `ProcessLambdaEventAsync` and `ProcessLambdaEventWithBatchResponseAsync`. But it's easier to iterate on tests now that the setup is in place, so open to feedback if you think we should add additional test cases.


test/AWS.Messaging.Tests.LambdaFunctions/Functions.cs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+public class SimulatedMessage
+{
+    public string? Id { get; set; }
+
+    public bool ReturnFailedStatus { get; set; } = false;
+
+    public TimeSpan WaitTime { get; set; } = TimeSpan.Zero;
+}
```
96malhar, line 79: How is this different from the `SimulatedMessage` created in the `AWS.Messaging.IntegrationTests.Models` namespace?

96malhar (changes requested):
Overall I am fine with the testing approach that we are taking. But I'd like to see 2 additional logical pieces being exercised (should be easy to add since the `AWSUtilities` and `TestUtilities` does most of the heavy lifting)

* Verify that the message source is correctly resolved from `AWS_LAMBDA_FUNCTION_NAME` environment variable
  * Log the message source to cloudwatch and assert on it.
* Verify FIFO processing of messages
  * Process 3 or more messages (belonging to the same group)  from a FIFO queue and log their `ProcessEnd` times to CloudWatch. Assert that they were processed in the same order in which they were published.

test/AWS.Messaging.Tests.LambdaFunctions/Functions.cs
ashovlin: It's the same content. I wanted to avoid a reference from `AWS.Messaging.Tests.LambdaFunctions` on `AWS.Messaging.IntegrationTests` because we were previously compiling `LambdaFunctions` during the compilation of `IntegrationTests`. But might be moot now that https://github.com/awslabs/aws-dotnet-messaging/pull/52#discussion_r1218779978 has been addressed, I'll take another look. Or maybe move both into `AWS.Messaging.Tests.Common`.

philasmar (approved):

test/AWS.Messaging.IntegrationTests/LambdaTests.cs
```diff
@@ -0,0 +1,339 @@
...
+{
+    public string ExecutionRoleArn { get; set; } = string.Empty;
+
+    public string ArtifactBucketName { get; set; } = string.Empty;
+
+    /// <summary>
+    /// This is both the name of the local package containing the Lambda function(s) under test,
+    /// as well as the name of the deployed Lambda function
+    /// </summary>
+    public  string FunctionPackageName => "AWS.Messaging.Tests.LambdaFunctions";
```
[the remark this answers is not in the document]
ashovlin: Done. I originally wasn't sure if we'd vary the package name based on the code under test, but ended up following the _"one package with multiple handlers"_ pattern we have in the Lambda repo.

ashovlin:
> Overall I am fine with the testing approach that we are taking. But I'd like to see 2 additional logical pieces being exercised (should be easy to add since the `AWSUtilities` and `TestUtilities` does most of the heavy lifting)
> 
>     * Verify that the message source is correctly resolved from `AWS_LAMBDA_FUNCTION_NAME` environment variable
>       
>       * Log the message source to cloudwatch and assert on it.
> 
>     * Verify FIFO processing of messages
>       
>       * Process 3 or more messages (belonging to the same group)  from a FIFO queue and log their `ProcessEnd` times to CloudWatch. Assert that they were processed in the same order in which they were published.

Added FIFO tests in https://github.com/awslabs/aws-dotnet-messaging/pull/71/commits/c416d617262ad4dea9ffd9ebee0905eb6e7fd0c3.

I started going down the path of testing the message source, but currently we're not actually sending any messages from the Lambda function. I started a "bidirectional" test where we'd message the Lambda and then it'd reply with a forwarded message that we can assert the source on the client. But we'd have to expand the IAM permissions given to the Lambda to allow publishing (with the queueURL), as well as run a poller on the client. Overall it was feeling like more effort and setup code/time than it was worth.

96malhar (approved):

PR #77, opened 2023-11-28
chore: Add beta suffix to the package version numbers

96malhar:
*Description of changes:*
This PR adds the `beta` suffix to the package version numbers. NuGet treats any version number suffix to denote a pre-release version and also reflects the same on the package homepage. 

Example - https://www.nuget.org/packages/Amazon.Lambda.Annotations/0.10.0-preview
image

See [here](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#pre-release-versions) to read more about pre-release versioning.


philasmar (approved):

ashovlin (approved):

ashovlin:
I know we've gone back and forth on this since it does limit visibility, but okay with it in this case since we're doing an earlier "beta" than our usual "dev preview".

Issue #90, opened 2023-12-21
Publishing to SNS using Topic from external service API response


raghumirajkar:
Hi,

We have a use case where one of the microservices wishes to publish to a SNS topic using the topic ARN (and a few more details like Role to assume and the entire message) from an external service response like an SQS message body or an Event Bridge message. The Topic ARN and other details are part of the SQS/EB message. For ease, we can assume we have a REST API which has all these details in the payload - 
```csharp
[ApiController]
public class DemoController { 
  [HttpPost]
  public IActionResult PublishMessage([FromBody] Dictionay<string, object?> payload) {
          var topicARN = // get topic Arn from payload
          var roleARN = // get Role Arn from payload
          var message =  // get message from payload
          ...
         // 1. STS Assume Role with role ARN provided in the payload
         // 2. Prepare SNS message content using message from payload
         // 3. Publish message to topic provided in payload (topicARN)
   }
}
```
How can we achieve this using AWS Messaging SDK? As seen in the example, the Topic ARN is provided when adding the SNS publisher during service bootstrap -
https://github.com/awslabs/aws-dotnet-messaging/blob/main/sampleapps/PublisherAPI/Program.cs#L21

madhub:
This is valid use case where we need to publish message to customer provided **SQS topic ARN** from a micro service . In multi tenanted deployment we need to store  multiple customers topicARN internally as configuration data & publish message to a topic .

ashovlin:
Currently when registering that we'll publish a type of message, we require the destination as well as the optional message type (if different from the name of the `OrderInfo` type).
```
builder.Services.AddAWSMessageBus(bus =>
{
    bus.AddSNSPublisher<OrderInfo>("arn:aws:sns:us-west-2:012345678910:MPF", "orderInfo");
}
```

For a scenario like you describe, is the destination _always_ determined at runtime? Is there anything like a "default" or "fallback" destination? I'm mulling over whether we should still register these types of messages in advance (to collect the optional message type, as well as anything else we expand that to in the future) vs. specifying everything at the point of publishing a single message.

```
builder.Services.AddAWSMessageBus(bus =>
{
   // Should we still require something like this? Or can that just be specified per-message
   
   // We still want to publish OrderInfo objects as "orderInfo" messages, 
   // but the destination will be set per message
    bus.AddSNSPublisher<OrderInfo>("orderInfo");
}
```

raghumirajkar:
[...]

PR #93, opened 2024-02-05
feat: Add support for specifying the message destination and AWS service client when publishing a specific message.

ashovlin:
*Issue #, if available:* https://github.com/awslabs/aws-dotnet-messaging/issues/90, DOTNET-7333

*Description of changes:* This adds support for configuring the destination and AWS client at the point of publishing. This may be useful for multi-tenant applications as described in #90 , where the destination of the message (and therefore the credentials required to publish it) aren't known until building the message.

For configuring the publisher without a destination in advance, I considered:
1. A new method, like `AddGenericSQSPublisher`, (or maybe `Destinationless`? `Unknown`?) 
    * **Why not?** Couldn't think of a good name
2. Leaving the signature as is, but requiring users to register a fake destination `AddSQSPublisher("WILL_OVERRIDE_LATER")`
    * **Why not?** Don't want to end up in a state where a user forgets to set the message-specific destination and we're sending requests to a bad/unknown endpoint.
3. Flipping our existing `string queueURL` parameter to `string? queueURL` so allowing `AddSQSPublisher(null)`
     * **Why ?** Seems like a balance between options 1 and 2 above.

### DIFF
--- a/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
@@ -18,7 +18,7 @@ public interface IMessageBusBuilder
     /// </summary>
     /// <param name="queueUrl">The SQS queue URL to publish the message to.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
-    IMessageBusBuilder AddSQSPublisher<TMessage>(string queueUrl, string? messageTypeIdentifier = null);
+    IMessageBusBuilder AddSQSPublisher<TMessage>(string? queueUrl, string? messageTypeIdentifier = null);
 
     /// <summary>
     /// Adds an SNS Publisher to the framework which will handle publishing
@@ -26,7 +26,7 @@ public interface IMessageBusBuilder
     /// </summary>
     /// <param name="topicUrl">The SNS topic URL to publish the message to.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
-    IMessageBusBuilder AddSNSPublisher<TMessage>(string topicUrl, string? messageTypeIdentifier = null);
+    IMessageBusBuilder AddSNSPublisher<TMessage>(string? topicUrl, string? messageTypeIdentifier = null);
 
     /// <summary>
     /// Adds an EventBridge Publisher to the framework which will handle publishing the defined message type to the specified EventBridge event bus name.
@@ -35,7 +35,7 @@ public interface IMessageBusBuilder
     /// <param name="eventBusName">The EventBridge event bus name or ARN where the message will be published.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
     /// <param name="options">Contains additional properties that can be set while configuring an EventBridge publisher</param>
-    IMessageBusBuilder AddEventBridgePublisher<TMessage>(string eventBusName, string? messageTypeIdentifier = null, EventBridgePublishOptions? options = null);
+    IMessageBusBuilder AddEventBridgePublisher<TMessage>(string? eventBusName, string? messageTypeIdentifier = null, EventBridgePublishOptions? options = null);
 
     /// <summary>
     /// Add a message handler for a given message type.
--- a/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
@@ -1,6 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using AWS.Messaging.Publishers.EventBridge;
+using AWS.Messaging.Publishers.SNS;
+using AWS.Messaging.Publishers.SQS;
 using AWS.Messaging.Serialization;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -16,7 +19,8 @@ public interface IMessageBusBuilder
     /// Adds an SQS Publisher to the framework which will handle publishing
     /// the defined message type to the specified SQS queues URL.
     /// </summary>
-    /// <param name="queueUrl">The SQS queue URL to publish the message to.</param>
+    /// <param name="queueUrl">The SQS queue URL to publish the message to. If the queue URL is null, a message-specific queue
+    /// URL must be specified on the <see cref="SQSOptions"/> when sending a message.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
     IMessageBusBuilder AddSQSPublisher<TMessage>(string? queueUrl, string? messageTypeIdentifier = null);
 
@@ -24,7 +28,8 @@ public interface IMessageBusBuilder
     /// Adds an SNS Publisher to the framework which will handle publishing
     /// the defined message type to the specified SNS topic URL.
     /// </summary>
-    /// <param name="topicUrl">The SNS topic URL to publish the message to.</param>
+    /// <param name="topicUrl">The SNS topic URL to publish the message to. If the topic URL is null, a message-specific
+    /// topic URL must be set on the <see cref="SNSOptions"/> when publishing a message.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
     IMessageBusBuilder AddSNSPublisher<TMessage>(string? topicUrl, string? messageTypeIdentifier = null);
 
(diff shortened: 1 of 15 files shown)


src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
```diff
@@ -84,9 +92,22 @@ public async Task PublishAsync<T>(T message, EventBridgeOptions? eventBridgeOpti
 
                 var messageBody = await _envelopeSerializer.SerializeAsync(messageEnvelope);
 
-                _logger.LogDebug("Sending the message of type '{MessageType}' to EventBridge. Publisher Endpoint: {Endpoint}", typeof(T), publisherEndpoint);
-                var request = CreatePutEventsRequest(publisherMapping, messageEnvelope.Source?.ToString(), messageBody, eventBridgeOptions);
-                await _eventBridgeClient.PutEventsAsync(request, token);
+                var client = _eventBridgeClient;
```
normj, line 95: If a users is only using the `OverrideClient` property then it seems possible there are not environment configuration that the `awsClientProvider.GetServiceClient` will be able to resolve the creation of the service client. The `AWSClientProvider` will throw a `FailedToFindAWSServiceClientException` exception for a service client the user didn't even want. 

I'm guessing the problem exists in all 3 publishers.

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
```diff
@@ -84,9 +92,22 @@ public async Task PublishAsync<T>(T message, EventBridgeOptions? eventBridgeOpti
...
-                var request = CreatePutEventsRequest(publisherMapping, messageEnvelope.Source?.ToString(), messageBody, eventBridgeOptions);
-                await _eventBridgeClient.PutEventsAsync(request, token);
+                var client = _eventBridgeClient;
+                if (eventBridgeOptions?.OverrideClient != null)
+                {
+                    // Use the user-provided client
+                    client = eventBridgeOptions.OverrideClient;
+
+                    // But still update the user agent to match the built-in client
+                    if (client is AmazonServiceClient)
```
normj, line 102: Unfortunately I don't think there is a safe way to add this callback because you need to add only if hasn't already been added to this instance. In your case if the same instance is used for `OverrideClient` then the `AWSServiceClient_BeforeServiceRequest` is added multiple times and called for each time it is added. This callback should generally only be added as part of service client creation not on a per operation situation.

normj (changes requested):

src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
```diff
@@ -18,15 +18,15 @@ public interface IMessageBusBuilder
     /// </summary>
     /// <param name="queueUrl">The SQS queue URL to publish the message to.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
-    IMessageBusBuilder AddSQSPublisher<TMessage>(string queueUrl, string? messageTypeIdentifier = null);
+    IMessageBusBuilder AddSQSPublisher<TMessage>(string? queueUrl, string? messageTypeIdentifier = null);
```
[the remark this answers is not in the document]
ashovlin: Expanded in https://github.com/awslabs/aws-dotnet-messaging/pull/93/commits/84282884e9c5f601bb85c2533e36b6f63d7bcf29

src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
```diff
@@ -18,15 +18,15 @@ public interface IMessageBusBuilder
...
 
     /// <summary>
     /// Adds an SNS Publisher to the framework which will handle publishing
     /// the defined message type to the specified SNS topic URL.
     /// </summary>
     /// <param name="topicUrl">The SNS topic URL to publish the message to.</param>
     /// <param name="messageTypeIdentifier">The language-agnostic message type identifier. If not specified, the .NET type will be used.</param>
-    IMessageBusBuilder AddSNSPublisher<TMessage>(string topicUrl, string? messageTypeIdentifier = null);
+    IMessageBusBuilder AddSNSPublisher<TMessage>(string? topicUrl, string? messageTypeIdentifier = null);
```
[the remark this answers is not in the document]
ashovlin: Expanded in https://github.com/awslabs/aws-dotnet-messaging/pull/93/commits/84282884e9c5f601bb85c2533e36b6f63d7bcf29

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
ashovlin: Removed the user agent handling after our internal discussion. We lose some future ability to filter these requests based on whether they're coming from the messaging framework vs. the underlying SDK, but we also want to be cautious around manipulating the user-provided clients.

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
ashovlin: Still working on this, but documenting current state. Since `MessageBusBuilder` is still calling `services.TryAddAWSService<Amazon.SQS.IAmazonSQS>()`, even for a destination-less publisher, I don't think we're actually triggering the `FailedToFindAWSServiceClientException`

With no `[default]` credentials or region at all is actually an exception from down in NetCore.Setup, not the `FailedToFindAWSServiceClientException`.
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
       ---> Amazon.Runtime.AmazonClientException: No RegionEndpoint or ServiceURL configured
         at Amazon.Runtime.ClientConfig.Validate()
         at Amazon.Runtime.AmazonServiceClient..ctor(AWSCredentials credentials, ClientConfig config)
         --- End of inner exception stack trace ---
         at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
         at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
         at Amazon.Extensions.NETCore.Setup.ClientFactory.CreateClient(Type serviceInterfaceType, AWSCredentials credentials, ClientConfig config)
         at Amazon.Extensions.NETCore.Setup.ClientFactory.CreateServiceClient(ILogger logger, Type serviceInterfaceType, AWSOptions options)
         at Amazon.Extensions.NETCore.Setup.ClientFactory.CreateServiceClient(IServiceProvider provider)
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
         at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(Type serviceType)
         at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
         at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
         at AWS.Messaging.Configuration.AWSClientProvider.GetServiceClient[T]() in C:\dev\aws-dotnet-messaging\src\AWS.Messaging\Configuration\AWSClientProvider.cs:line 31
         at AWS.Messaging.Publishers.SQS.SQSPublisher..ctor(IAWSClientProvider awsClientProvider, ILogger`1 logger, IMessageConfiguration messageConfiguration, IEnvelopeSerializer envelopeSerializer, ITelemetryFactory telemetryFactory) in C:\dev\aws-dotnet-messaging\src\AWS.Messaging\Publishers\SQS\SQSPublisher.cs:line 38
         at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean 
```

With a region but no credentials then we get this from in the SDK
```
Amazon.Runtime.AmazonServiceException: Unable to get IAM security credentials from EC2 Instance Metadata Service.
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.FetchCredentials()
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentials()
   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentialsAsync()
    ...
```

normj (approved):

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
```diff
@@ -97,21 +121,22 @@ public async Task PublishAsync<T>(T message, EventBridgeOptions? eventBridgeOpti
...
-            EventBusName = publisherEndpoint,
+            EventBusName = eventBusName,
             DetailType = publisherMapping.MessageTypeIdentifier,
             Detail = messageBody
         };
 
         var putEventsRequest = new PutEventsRequest
         {
-            EndpointId = publisherConfiguration.EndpointID,
+            // Give precendence to the endpoint ID if specified on the message-specific eventBridgeOptions,
```
philasmar, line 137: nit: typo in "precendence"

philasmar (approved):

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
ashovlin: Fixed in https://github.com/awslabs/aws-dotnet-messaging/pull/93/commits/f63889552d886d188961ae759047f5a372447e48

PR #95, opened 2024-02-09
feat: add ability for users to enable data message logging

philasmar:
*Issue #, if available:*
DOTNET-7340
DOTNET-7342

*Description of changes:*
* Added a new method `EnableDataMessageLogging` in the message bus builder that would enable data messages to be logged as well as appear in any exceptions thrown.
* Added tests, updated the sample apps and README

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,14 @@ We welcome community contributions and pull requests. See [CONTRIBUTING.md](./CO
 # Security
 The AWS Message Processing Framework for .NET relies on the [AWS SDK for .NET](https://github.com/aws/aws-sdk-net) for communicating with AWS. Refer to the [security section](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/security.html) in the [AWS SDK for .NET Developer Guide](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/welcome.html) for more information.
 
+The framework does not log data messages sent by the user for security purposes. If users want to enabled this functionality for debugging purposes, you need to call `EnableDataMessageLogging()` in the AWS Message Bus as follows:
+```csharp
+builder.Services.AddAWSMessageBus(bus =>
+{
+    builder.EnableDataMessageLogging();
+});
+```
+
 If you discover a potential security issue, refer to the [security policy](https://github.com/awslabs/aws-dotnet-messaging/security/policy) for reporting information.
 
 # Additional Resources
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ We welcome community contributions and pull requests. See [CONTRIBUTING.md](./CO
 # Security
 The AWS Message Processing Framework for .NET relies on the [AWS SDK for .NET](https://github.com/aws/aws-sdk-net) for communicating with AWS. Refer to the [security section](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/security.html) in the [AWS SDK for .NET Developer Guide](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/welcome.html) for more information.
 
-The framework does not log data messages sent by the user for security purposes. If users want to enabled this functionality for debugging purposes, you need to call `EnableDataMessageLogging()` in the AWS Message Bus as follows:
+The framework does not log data messages sent by the user for security purposes. If users want to enable this functionality for debugging purposes, you need to call `EnableDataMessageLogging()` in the AWS Message Bus as follows:
 ```csharp
 builder.Services.AddAWSMessageBus(bus =>
 {
--- a/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
+++ b/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
@@ -73,5 +73,5 @@ public interface IMessageConfiguration
     /// If this is enabled, messages sent by this framework will be visible in plain text across the framework's components.
     /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
     /// </summary>
-    bool DataMessageLogging { get; set; }
+    bool LogMessageContent { get; set; }
 }
--- a/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
+++ b/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
@@ -67,4 +67,11 @@ public interface IMessageConfiguration
     /// computed by the framework in the absence of a user-defined one.
     /// </summary>
     string? SourceSuffix { get; set; }
+
+    /// <summary>
+    /// Controls the visibility of data messages in the logging framework, exception handling and other areas.
+    /// If this is enabled, messages sent by this framework will be visible in plain text across the framework's components.
+    /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
+    /// </summary>
+    bool DataMessageLogging { get; set; }
 }
--- a/src/AWS.Messaging/Configuration/Internal/ApplicationSettings.cs
+++ b/src/AWS.Messaging/Configuration/Internal/ApplicationSettings.cs
@@ -40,6 +40,13 @@ public class ApplicationSettings
     public ICollection<SQSPoller> SQSPollers { get; set; } = default!;
 
     /// <summary>
+    /// Controls the visibility of data messages in the logging framework, exception handling and other areas.
+    /// If this is enabled, messages sent by this framework will be visible in plain text across the framework's components.
+    /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
+    /// </summary>
+    public bool? LogMessageContent { get; set; } = default;
+
+    /// <summary>
     /// Represents an SQS publisher configuration.
     /// </summary>
     public class SQSPublisher
--- a/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
@@ -239,6 +239,9 @@ public class MessageBusBuilder : IMessageBusBuilder
             }
         }
 
+        if (settings.LogMessageContent != null)
+            _messageConfiguration.LogMessageContent = settings.LogMessageContent ?? false;
+
         return this;
     }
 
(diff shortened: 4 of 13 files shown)


README.md
```diff
@@ -311,6 +311,14 @@ We welcome community contributions and pull requests. See [CONTRIBUTING.md](./CO
 # Security
 The AWS Message Processing Framework for .NET relies on the [AWS SDK for .NET](https://github.com/aws/aws-sdk-net) for communicating with AWS. Refer to the [security section](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/security.html) in the [AWS SDK for .NET Developer Guide](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/welcome.html) for more information.
 
+The framework does not log data messages sent by the user for security purposes. If users want to enabled this functionality for debugging purposes, you need to call `EnableDataMessageLogging()` in the AWS Message Bus as follows:
+```csharp
+builder.Services.AddAWSMessageBus(bus =>
+{
+    builder.EnableDataMessageLogging();
```
normj, line 318: We should have a way for this to be enabled for `IConfiguration` so users can enable this without changing code.

src/AWS.Messaging/Configuration/IMessageConfiguration.cs
```diff
@@ -67,4 +67,11 @@ public interface IMessageConfiguration
     /// computed by the framework in the absence of a user-defined one.
     /// </summary>
     string? SourceSuffix { get; set; }
+
+    /// <summary>
+    /// Controls the visibility of data messages in the logging framework, exception handling and other areas.
+    /// If this is enabled, messages sent by this framework will be visible in plain text across the framework's components.
+    /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
+    /// </summary>
+    bool DataMessageLogging { get; set; }
```
normj, line 76: My preference for naming is `LogMessageContent`.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -163,6 +175,11 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
             _logger.LogTrace("Created a generic {MessageEnvelopeName} of type '{MessageEnvelopeType}'", nameof(MessageEnvelope), result.Envelope.GetType());
             return result;
         }
+        catch (JsonException) when (!_messageConfiguration.DataMessageLogging)
```
normj, line 178: Is the idea that if the message is `JsonException` the inner exception might contain the message body? This seems hard to debug if we don't give the user any of the JSON parsing error message.

src/AWS.Messaging/Serialization/MessageSerializer.cs
```diff
@@ -29,9 +29,22 @@ public object Deserialize(string message, Type deserializedType)
...
+                _logger.LogTrace("Deserializing the following message into type '{DeserializedType}':\n{Message}", deserializedType, message);
+            }
+            else
+            {
+                _logger.LogTrace("Deserializing the following message into type '{DeserializedType}'", deserializedType);
+            }
+
             return JsonSerializer.Deserialize(message, deserializedType, jsonSerializerOptions) ?? throw new JsonException("The deserialized application message is null.");
         }
+        catch (JsonException) when (!_messageConfiguration.DataMessageLogging)
```
normj, line 43: Same question about not giving any information about the JSON parse error.

src/AWS.Messaging/Serialization/MessageSerializer.cs
```diff
@@ -47,9 +60,22 @@ public string Serialize(object message)
...
+                _logger.LogTrace("Serialized the message object as the following raw string:\n{JsonString}", jsonString);
+            }
+            else
+            {
+                _logger.LogTrace("Serialized the message object to a raw string");
+            }
+
             return jsonString;
         }
+        catch (JsonException) when (!_messageConfiguration.DataMessageLogging)
```
normj, line 74: Same question about not giving any information about the JSON parse error.

src/AWS.Messaging/Serialization/MessageSerializer.cs
```diff
@@ -47,9 +60,22 @@ public string Serialize(object message)
...
             var jsonSerializerOptions = _messageConfiguration.SerializationOptions.SystemTextJsonOptions;
             var jsonString = JsonSerializer.Serialize(message, jsonSerializerOptions);
-            _logger.LogTrace("Serialized the message object as the following raw string:\n{JsonString}", jsonString);
+            if (_messageConfiguration.DataMessageLogging)
+            {
+                _logger.LogTrace("Serialized the message object as the following raw string:\n{JsonString}", jsonString);
+            }
+            else
+            {
+                _logger.LogTrace("Serialized the message object to a raw string");
```
normj, line 69: Maybe given them the size of the raw string to they know it is not an empty string or something?

normj (changes requested):

README.md
philasmar: Done

src/AWS.Messaging/Configuration/IMessageConfiguration.cs
philasmar: Done

src/AWS.Messaging/Serialization/MessageSerializer.cs
philasmar: Done

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
philasmar: Yes correct. The PenTester said since we are unsure what content might/will be exposed in the inner exception, then we need to remove or redact this information. The messages I have seen from JsonException are not particularly the most helpful exceptions.

src/AWS.Messaging/Serialization/MessageSerializer.cs
philasmar: Yes correct. The PenTester said since we are unsure what content might/will be exposed in the inner exception, then we need to remove or redact this information. The messages I have seen from JsonException are not particularly the most helpful exceptions.

src/AWS.Messaging/Serialization/MessageSerializer.cs
philasmar: Yes correct. The PenTester said since we are unsure what content might/will be exposed in the inner exception, then we need to remove or redact this information. The messages I have seen from JsonException are not particularly the most helpful exceptions.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -239,6 +239,9 @@ public IMessageBusBuilder LoadConfigurationFromSettings(IConfiguration configura
             }
         }
 
+        if (settings.LogMessageContent != null)
+            _messageConfiguration.LogMessageContent = settings.LogMessageContent ?? false;
```
normj, line 243: This only works if a user is using the `LoadConfigurationFromSettings` method. I think a user should always be able to turn this on without having to change code. Or maybe `LoadConfigurationFromSettings` should always be code and if there is no mapping in `IConfiguration` it wouldn't do anything.

normj (changes requested):

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: I have added a `LoadConfigurationFromEnvironment` which will be called automatically in the builder. The reason why I did not call `LoadConfigurationFromSettings` automatically (which would've also tackled our issue) is because that method needs access to the IConfiguration service which is not accessible automatically in the MessageBusBuilder.

README.md
```diff
@@ -311,6 +311,14 @@ We welcome community contributions and pull requests. See [CONTRIBUTING.md](./CO
 # Security
 The AWS Message Processing Framework for .NET relies on the [AWS SDK for .NET](https://github.com/aws/aws-sdk-net) for communicating with AWS. Refer to the [security section](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/security.html) in the [AWS SDK for .NET Developer Guide](https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/welcome.html) for more information.
 
+The framework does not log data messages sent by the user for security purposes. If users want to enabled this functionality for debugging purposes, you need to call `EnableDataMessageLogging()` in the AWS Message Bus as follows:
```
96malhar, line 314: nit: "enabled" -> "enable"

src/AWS.Messaging/Configuration/Internal/ApplicationSettings.cs
```diff
@@ -39,6 +39,13 @@ public class ApplicationSettings
     /// </summary>
     public ICollection<SQSPoller> SQSPollers { get; set; } = default!;
 
+    /// <summary>
+    /// Controls the visibility of data messages in the logging framework, exception handling and other areas.
+    /// If this is enabled, messages sent by this framework will be visible in plain text across the framework's components.
+    /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
+    /// </summary>
+    public bool? LogMessageContent { get; set; } = default;
```
96malhar, line 47: Should this be nullable? In the `MessageBusBuilder` you are always defaulting to `false` if this value is null.

96malhar: What I meant was that `LogMessageContent ` should be an non-nullable `bool` in `ApplicationSettings ` as well. That way if a user does not specify any value for it in the `appsettings.json` file, then we end up with an explicit `false` and we'll avoid the null check while setting this property on the `_messageConfiguration`.

image

Anyway, this is minor and not a deal breaker.

README.md
philasmar: Fixed

normj (approved):

96malhar (approved):

src/AWS.Messaging/Configuration/Internal/ApplicationSettings.cs
philasmar: Them not specifying a value for it is not indicative of what the value should be as there are multiple sources for this value. I wanted ApplicationSettings to be indicative of what values were provided.

PR #96, opened 2024-02-14
chore: add validation to the ECS Task Metadata

philasmar:
*Issue #, if available:*
DOTNET-7341

*Description of changes:*
* Added validation for the ECS Metadata endpoint to look for the IP `169.254.170.2`
* Added a POCO that represents the data returned by the endpoint which adds more validation on the structure of the data
* Added validation to the Cluster and TaskARN returned

### DIFF
--- a/src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
+++ b/src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
@@ -47,18 +47,26 @@ internal class ECSContainerMetadataManager : IECSContainerMetadataManager
 
             var response = await client.GetAsync(new Uri($"{ecsMetadataUri}/task"));
             if (!response.IsSuccessStatusCode)
+            {
+                _logger.LogError("Unable to retrieve task metadata from the ECS container.");
                 return null;
+            }
 
             var taskMetadataJson = await response.Content.ReadAsStringAsync();
             var taskMetadata = JsonSerializer.Deserialize<TaskMetadataResponse>(taskMetadataJson);
-            return
-                ValidateContainerTaskMetadata(taskMetadata) ?
-                    taskMetadata :
-                    null;
+            if (ValidateContainerTaskMetadata(taskMetadata))
+            {
+                return taskMetadata;
+            }
+            else
+            {
+                _logger.LogError("The retrieved task metadata from the ECS container is invalid.");
+                return null;
+            }
         }
         catch (Exception ex)
         {
-            _logger.LogError(ex, "Unable to retrieve Task Arn from ECS container metadata.");
+            _logger.LogError(ex, "Unable to retrieve task metadata from the ECS container.");
             return null;
         }
     }
--- a/src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
+++ b/src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
@@ -2,6 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Text.Json;
+using System.Text.RegularExpressions;
+using Amazon;
+using AWS.Messaging.Configuration.Internal;
 using Microsoft.Extensions.Logging;
 
 namespace AWS.Messaging.Services;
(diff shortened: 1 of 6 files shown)


src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
```diff
@@ -13,43 +16,86 @@ internal class ECSContainerMetadataManager : IECSContainerMetadataManager
...
+            if (string.IsNullOrEmpty(metadata.Cluster))
+                return false;
+
+            // Check if the string length is within the allowed limit (255 characters)
+            if (metadata.Cluster.Length > 255)
+                return false;
+
+            // Use a regular expression to validate the characters
+            // Valid characters include letters (uppercase and lowercase), numbers, hyphens, and underscores
+            Regex validPattern = new Regex(@"^[A-Za-z0-9\-_]+$");
```
philasmar, line 90: I am on the fence about the use of RegEx and it's performance impact in this context. Any thoughts?

normj: If I understand this should only ever be called once so it should have negligible performance impact. Also since it should use the `Compiled` option since it is only being used. Not that you are using it but adding that feedback in case another reviewer suggest it.

normj (approved):

src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
```diff
@@ -13,43 +16,86 @@ internal class ECSContainerMetadataManager : IECSContainerMetadataManager
...
+                return null;
 
             var taskMetadataJson = await response.Content.ReadAsStringAsync();
-            return JsonSerializer.Deserialize<Dictionary<string, object>>(taskMetadataJson) ??
-                metadata;
+            var taskMetadata = JsonSerializer.Deserialize<TaskMetadataResponse>(taskMetadataJson);
+            return
+                ValidateContainerTaskMetadata(taskMetadata) ?
+                    taskMetadata :
+                    null;
```
96malhar, line 57: We should log an error here saying _"Encountered invalid ECS task metadata..."_ before returning `null`.

src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
```diff
@@ -13,43 +16,86 @@ internal class ECSContainerMetadataManager : IECSContainerMetadataManager
...
 
         try
         {
             var client = _httpClientFactory.CreateClient("ECSMetadataClient");
 
-            var response = await client.GetAsync(new Uri($"{ecsMetadataURI}/task"));
+            var response = await client.GetAsync(new Uri($"{ecsMetadataUri}/task"));
             if (!response.IsSuccessStatusCode)
-                return metadata;
+                return null;
```
96malhar, line 50: We should log an error here saying _"Unable to retrieve ECS task metadata..."_ before returning null.

96malhar (changes requested):

src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
philasmar: Fixed

src/AWS.Messaging/Services/ECSContainerMetadataManager.cs
philasmar: Fixed

test/AWS.Messaging.UnitTests/ECSContainerMetadataManagerTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            _httpClientFactory.Object,
+            _logger.Object);
+
+        var metadata = await ecsContainerMetadataManager.GetContainerTaskMetadata();
+
+        Assert.Null(metadata);
+    }
+
+    [Fact]
+    public async Task GetContainerTaskMetadata_InvalidJsonReturned()
```
96malhar, line 87: This test would have to be amended since we added additional log messages when then task metadata is missing or invalid.

philasmar: Fixed

96malhar (approved):

PR #107, opened 2024-03-13
feat: add backoff logic to the SQS Poller

philasmar:
*Issue #, if available:*
DOTNET-7390

*Description of changes:*
* Create interfaces for IBackoffHandler and IBackoffPolicy
* Add message bus configuration points for specifying the backoff policy, which for now is no backoff, interval backoff or capped exponential backoff
* Implement the 3 backoff policies: No backoff, Interval backoff and Capped Exponential backoff
* Implement the default backoff handler
* Add a sample app that replaces our BackoffHandler with Polly
* Add tests and update docs

### DIFF
--- a/sampleapps/SubscriberService/Program.cs
+++ b/sampleapps/SubscriberService/Program.cs
@@ -2,9 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Text.Json;
+using Amazon;
+using Amazon.Extensions.NETCore.Setup;
+using Amazon.Runtime;
+using Amazon.SQS;
 using AWS.Messaging.Telemetry.OpenTelemetry;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Resources;
@@ -24,6 +29,16 @@ await Host.CreateDefaultBuilder(args)
     })
     .ConfigureServices((context, services) =>
     {
+        AWSConfigs.LoggingConfig.LogTo = LoggingOptions.Console;
+        AWSConfigs.LoggingConfig.LogMetricsFormat = LogMetricsFormatOption.JSON;
+        AWSConfigs.LoggingConfig.LogResponses = ResponseLoggingOption.Always;
+        AWSConfigs.LoggingConfig.LogMetrics = true;
+
+        var sqsClient = new AmazonSQSClient(new AmazonSQSConfig
+        {
+            MaxErrorRetry = 0
+        });
+        services.TryAddSingleton<IAmazonSQS>(sqsClient);
         services.AddAWSMessageBus(builder =>
         {
             // To load the configuration from appsettings.json instead of the code below, uncomment this and remove the following lines.
@@ -32,22 +47,18 @@ await Host.CreateDefaultBuilder(args)
             builder.AddSQSPoller("https://sqs.us-west-2.amazonaws.com/012345678910/MPF");
             builder.AddMessageHandler<ChatMessageHandler, ChatMessage>("chatMessage");
 
-            builder.ConfigureSerializationOptions(options =>
+            // Optional: Configure the backoff policy used by the SQS Poller.
+            builder.ConfigureBackoffPolicy(options =>
             {
-                options.SystemTextJsonOptions = new JsonSerializerOptions
+                options.UseCappedExponentialBackoff(x =>
                 {
-                    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
-                };
+                    x.CapBackoffTime = 3600000;
+                });
             });
 
             // Logging data messages is disabled by default to protect sensitive user data. If you want this enabled, uncomment the line below.
-            // builder.EnableDataMessageLogging();
-        })
-        .AddOpenTelemetry()
-            .ConfigureResource(resource => resource.AddService("SubscriberService"))
-            .WithTracing(tracing => tracing
-                .AddAWSMessagingInstrumentation()
-                .AddConsoleExporter());
+            // builder.EnableMessageContentLogging();
+        });
     })
     .Build()
     .RunAsync();
--- a/sampleapps/SubscriberService/Program.cs
+++ b/sampleapps/SubscriberService/Program.cs
@@ -45,7 +45,7 @@ await Host.CreateDefaultBuilder(args)
             {
                 options.UseCappedExponentialBackoff(x =>
                 {
-                    x.CapBackoffTime = 3600;
+                    x.CapBackoffTime = 60;
                 });
             });
 
--- a/sampleapps/SubscriberService/Program.cs
+++ b/sampleapps/SubscriberService/Program.cs
@@ -2,14 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Text.Json;
-using Amazon;
-using Amazon.Extensions.NETCore.Setup;
-using Amazon.Runtime;
-using Amazon.SQS;
 using AWS.Messaging.Telemetry.OpenTelemetry;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Resources;
@@ -29,16 +24,6 @@ await Host.CreateDefaultBuilder(args)
     })
     .ConfigureServices((context, services) =>
     {
-        AWSConfigs.LoggingConfig.LogTo = LoggingOptions.Console;
-        AWSConfigs.LoggingConfig.LogMetricsFormat = LogMetricsFormatOption.JSON;
-        AWSConfigs.LoggingConfig.LogResponses = ResponseLoggingOption.Always;
-        AWSConfigs.LoggingConfig.LogMetrics = true;
-
-        var sqsClient = new AmazonSQSClient(new AmazonSQSConfig
-        {
-            MaxErrorRetry = 0
-        });
-        services.TryAddSingleton<IAmazonSQS>(sqsClient);
         services.AddAWSMessageBus(builder =>
         {
             // To load the configuration from appsettings.json instead of the code below, uncomment this and remove the following lines.
@@ -47,18 +32,31 @@ await Host.CreateDefaultBuilder(args)
             builder.AddSQSPoller("https://sqs.us-west-2.amazonaws.com/012345678910/MPF");
             builder.AddMessageHandler<ChatMessageHandler, ChatMessage>("chatMessage");
 
+            builder.ConfigureSerializationOptions(options =>
+            {
+                options.SystemTextJsonOptions = new JsonSerializerOptions
+                {
+                    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                };
+            });
+
             // Optional: Configure the backoff policy used by the SQS Poller.
             builder.ConfigureBackoffPolicy(options =>
             {
                 options.UseCappedExponentialBackoff(x =>
                 {
-                    x.CapBackoffTime = 3600000;
+                    x.CapBackoffTime = 3600;
                 });
             });
 
             // Logging data messages is disabled by default to protect sensitive user data. If you want this enabled, uncomment the line below.
             // builder.EnableMessageContentLogging();
-        });
+        })
+        .AddOpenTelemetry()
+            .ConfigureResource(resource => resource.AddService("SubscriberService"))
+            .WithTracing(tracing => tracing
+                .AddAWSMessagingInstrumentation()
+                .AddConsoleExporter());
     })
     .Build()
     .RunAsync();
--- a/src/AWS.Messaging/Configuration/BackoffPolicyBuilder.cs
+++ b/src/AWS.Messaging/Configuration/BackoffPolicyBuilder.cs
@@ -1,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using AWS.Messaging.Configuration.Internal;
+using AWS.Messaging.Services.Backoff.Policies.Options;
+
+namespace AWS.Messaging.Configuration;
+
+/// <summary>
+/// This builder is used to configure the backoff policy and it's options.
+/// </summary>
+public class BackoffPolicyBuilder : IBackoffPolicyBuilder
+{
+    private readonly MessageConfiguration _messageConfiguration;
+
+    /// <summary>
+    /// Creates an instance of <see cref="BackoffPolicyBuilder"/>.
+    /// </summary>
+    public BackoffPolicyBuilder(MessageConfiguration messageConfiguration)
+    {
+        _messageConfiguration = messageConfiguration;
+    }
+
+    /// <inheritdoc/>
+    public void UseNoBackoff()
+    {
+        _messageConfiguration.BackoffPolicy = BackoffPolicy.None;
+    }
+
+    /// <inheritdoc/>
+    public void UseIntervalBackoff(Action<IntervalBackoffOptions>? options = null)
+    {
+        _messageConfiguration.BackoffPolicy = BackoffPolicy.Interval;
+        options?.Invoke(_messageConfiguration.IntervalBackoffOptions);
+
+        _messageConfiguration.IntervalBackoffOptions.Validate();
+    }
+
+    /// <inheritdoc/>
+    public void UseCappedExponentialBackoff(Action<CappedExponentialBackoffOptions>? options = null)
+    {
+        _messageConfiguration.BackoffPolicy = BackoffPolicy.CappedExponential;
+        options?.Invoke(_messageConfiguration.CappedExponentialBackoffOptions);
+
+        _messageConfiguration.CappedExponentialBackoffOptions.Validate();
+    }
+}
--- a/src/AWS.Messaging/Configuration/BackoffPolicyBuilder.cs
+++ b/src/AWS.Messaging/Configuration/BackoffPolicyBuilder.cs
@@ -7,7 +7,7 @@ using AWS.Messaging.Services.Backoff.Policies.Options;
 namespace AWS.Messaging.Configuration;
 
 /// <summary>
-/// This builder is used to configure the backoff policy and it's options.
+/// This builder is used to configure the backoff policy and its options.
 /// </summary>
 public class BackoffPolicyBuilder : IBackoffPolicyBuilder
 {
--- a/src/AWS.Messaging/Configuration/IBackoffPolicyBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IBackoffPolicyBuilder.cs
@@ -1,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using AWS.Messaging.Services.Backoff.Policies;
+using AWS.Messaging.Services.Backoff.Policies.Options;
+
+namespace AWS.Messaging.Configuration;
+
+/// <summary>
+/// This builder interface is used to configure the backoff policy and it's options.
+/// </summary>
+public interface IBackoffPolicyBuilder
+{
+    /// <summary>
+    /// Sets the backoff policy to <see cref="NoBackoffPolicy"/>, effectively disabling back-offs.
+    /// </summary>
+    void UseNoBackoff();
+
+    /// <summary>
+    /// Sets the backoff policy to <see cref="IntervalBackoffPolicy"/> and allows it's configuration.
+    /// </summary>
+    void UseIntervalBackoff(Action<IntervalBackoffOptions> options);
+
+    /// <summary>
+    /// Sets the backoff policy to <see cref="CappedExponentialBackoffPolicy"/> and allows it's configuration.
+    /// </summary>
+    void UseCappedExponentialBackoff(Action<CappedExponentialBackoffOptions> options);
+}
--- a/src/AWS.Messaging/Configuration/IBackoffPolicyBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IBackoffPolicyBuilder.cs
@@ -7,7 +7,7 @@ using AWS.Messaging.Services.Backoff.Policies.Options;
 namespace AWS.Messaging.Configuration;
 
 /// <summary>
-/// This builder interface is used to configure the backoff policy and it's options.
+/// This builder interface is used to configure the backoff policy and its options.
 /// </summary>
 public interface IBackoffPolicyBuilder
 {
@@ -17,12 +17,12 @@ public interface IBackoffPolicyBuilder
     void UseNoBackoff();
 
     /// <summary>
-    /// Sets the backoff policy to <see cref="IntervalBackoffPolicy"/> and allows it's configuration.
+    /// Sets the backoff policy to <see cref="IntervalBackoffPolicy"/> and allows its configuration.
     /// </summary>
     void UseIntervalBackoff(Action<IntervalBackoffOptions> options);
 
     /// <summary>
-    /// Sets the backoff policy to <see cref="CappedExponentialBackoffPolicy"/> and allows it's configuration.
+    /// Sets the backoff policy to <see cref="CappedExponentialBackoffPolicy"/> and allows its configuration.
     /// </summary>
     void UseCappedExponentialBackoff(Action<CappedExponentialBackoffOptions> options);
 }
--- a/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
@@ -5,6 +5,7 @@ using AWS.Messaging.Publishers.EventBridge;
 using AWS.Messaging.Publishers.SNS;
 using AWS.Messaging.Publishers.SQS;
 using AWS.Messaging.Serialization;
+using AWS.Messaging.Services.Backoff;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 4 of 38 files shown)


src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+            ExceptionDispatchInfo? capturedException = null;
+            shouldRetry = false;
+
+            try
+            {
+                await task.Invoke();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "We have encountered an issue while polling '{SubscriberEndpoint}'.", configuration.SubscriberEndpoint);
```
ashovlin, line 55: nit: Don't love `"we`" since it might be unclear who it is (the application? the framework? the service?)

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+            else
+            {
+                // Checking the backoff policy for how long to backoff before attempting to poll SQS for messages again.
+                var waitTime = _backoffPolicy.RetrieveBackoffTime(retries);
+                _logger.LogWarning("Backing off for {WaitTime}ms before trying again...", waitTime);
+
+                await Task.Delay(TimeSpan.FromMilliseconds(waitTime), token);
+
+                retries++;
+                _logger.LogWarning("Attempting to poll SQS for messages once again...");
```
ashovlin, line 77: Could we log the retry count here? It's something that jumps out to me when working the SDK's retries
> "Attempt #{0} to poll for SQS messages"

src/AWS.Messaging/Services/Backoff/Policies/IntervalBackoffPolicy.cs
```diff
@@ -0,0 +1,47 @@
...
+    public bool ShouldBackoff(Exception exception, SQSMessagePollerConfiguration configuration)
+    {
+        return !configuration.IsExceptionFatal(exception);
+    }
+
+    /// <summary>
+    /// Retrieves the backoff time which is a fixed interval, regardless of how many retries have already been performed.
+    /// </summary>
+    /// <param name="numberOfRetries">The number of times the <see cref="BackoffHandler"/> has retried a request after performing a backoff.</param>
+    /// <returns>A fixed interval representing the backoff time.</returns>
```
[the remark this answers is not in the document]
philasmar: The interfaces don't care and the implementer can go with any time unit. However, our implementations do care, since the BackoffHandler is doing a TimeSpan.FromMilliseconds(waitTime). I will switch to using seconds all over and I will add docs about the unit we are returning where it makes sense.

src/AWS.Messaging/Services/Backoff/IBackoffHandler.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using AWS.Messaging.Configuration;
+
+namespace AWS.Messaging.Services.Backoff;
+
+/// <summary>
+/// Interface for a backoff handler that is responsible for performing back-offs in the event that a given delegate throws an exception.
```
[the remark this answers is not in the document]
philasmar: In general, I used "backoff" when it is before "handler" or "policy". I used "back off" for the verb. "back-off" was used for plural because Jetbrains Rider kept complaining about "backoffs" being a typo so I switched to "back-offs". I will keep it as is for now unless you feel strongly about it.

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+        bool shouldRetry;
+        var retries = 0;
+        do
+        {
+            ExceptionDispatchInfo? capturedException = null;
+            shouldRetry = false;
+
+            try
+            {
+                await task.Invoke();
```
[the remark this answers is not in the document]
philasmar: The token is still being passed to the SDK call. If you refer to the `SQSMessagePoller` where we are passing the delegate to the `BackoffHandler`, the SQS call still takes in the cancellation token.
```
var receiveMessageResponse = await _sqsClient.ReceiveMessageAsync(receiveMessageRequest, token);
```

sampleapps/SubscriberService/Program.cs
```diff
@@ -24,6 +29,16 @@ await Host.CreateDefaultBuilder(args)
...
     {
+        AWSConfigs.LoggingConfig.LogTo = LoggingOptions.Console;
+        AWSConfigs.LoggingConfig.LogMetricsFormat = LogMetricsFormatOption.JSON;
+        AWSConfigs.LoggingConfig.LogResponses = ResponseLoggingOption.Always;
+        AWSConfigs.LoggingConfig.LogMetrics = true;
+
+        var sqsClient = new AmazonSQSClient(new AmazonSQSConfig
+        {
+            MaxErrorRetry = 0
+        });
```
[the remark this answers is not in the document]
philasmar: Good catch. I didn't mean to leave those in. I removed them.

src/AWS.Messaging/Configuration/BackoffPolicyBuilder.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using AWS.Messaging.Configuration.Internal;
+using AWS.Messaging.Services.Backoff.Policies.Options;
+
+namespace AWS.Messaging.Configuration;
+
+/// <summary>
+/// This builder is used to configure the backoff policy and it's options.
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Messaging/Configuration/IBackoffPolicyBuilder.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+using AWS.Messaging.Services.Backoff.Policies;
+using AWS.Messaging.Services.Backoff.Policies.Options;
+
+namespace AWS.Messaging.Configuration;
+
+/// <summary>
+/// This builder interface is used to configure the backoff policy and it's options.
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
```diff
@@ -103,4 +104,9 @@ public interface IMessageBusBuilder
     /// This means any sensitive user data sent by this framework will be visible in logs, any exceptions thrown and others.
     /// </summary>
     IMessageBusBuilder EnableMessageContentLogging();
+
+    /// <summary>
+    /// Configures the backoff policy used by <see cref="BackoffHandler"/> and it's available options.
```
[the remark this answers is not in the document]
philasmar: Fixed

src/AWS.Messaging/Services/Backoff/Policies/Options/IntervalBackoffOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+/// <summary>
+/// Configuration for the interval backoff policy.
+/// </summary>
+public class IntervalBackoffOptions
+{
+    /// <summary>
+    /// The fixed interval in milliseconds to wait between back-offs.
+    /// The default value is 1 second.
+    /// </summary>
+    public int FixedInterval { get; set; } = 1000;
```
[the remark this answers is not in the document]
philasmar: I used ms initially because I thought users might want to use sub-second intervals. Though I do agree that its better for consistency for us to use seconds. I changed it for now, but open to feedback from others.

src/AWS.Messaging/Services/Backoff/Policies/IntervalBackoffPolicy.cs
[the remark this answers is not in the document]
philasmar: Done

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
philasmar: Reverted.

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
philasmar: Done

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -242,6 +246,21 @@ public IMessageBusBuilder LoadConfigurationFromSettings(IConfiguration configura
         if (settings.LogMessageContent != null)
             _messageConfiguration.LogMessageContent = settings.LogMessageContent ?? false;
 
+        if (settings.BackoffPolicy != null)
+        {
+            _messageConfiguration.BackoffPolicy = settings.BackoffPolicy ?? BackoffPolicy.CappedExponential;
```
96malhar, line 251: You are already checking `settings.BackoffPolicy` for null in the `if` statement above. Not sure if the null-coalescing operator (??) serves any purpose here.

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (!shouldRetry)
+            {
+                capturedException.Throw();
+            }
+            else
+            {
+                // Checking the backoff policy for how long to backoff before attempting to poll SQS for messages again.
+                var waitTime = _backoffPolicy.RetrieveBackoffTime(retries);
+                _logger.LogWarning("Backing off for {WaitTime}s before trying again...", waitTime);
```
96malhar, line 72: I think it's worth being explicit in the log message that we are backing off before try to poll messages from SQS. This will help in identifying the action that we are backing off from in-case we add retries to other pieces in MPF

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace AWS.Messaging.Services.Backoff.Policies.Options;
+
+/// <summary>
+/// Configuration for the capped exponential backoff policy.
+/// </summary>
+public class CappedExponentialBackoffOptions
+{
+    /// <summary>
+    /// The backoff time in seconds to cap the exponential backoff at.
```
96malhar, line 12: This should also mention that once this threshold is reached, we will always perform retries at a fixed interval of `CapBackoffTime`.

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Validates that the options set for the <see cref="CappedExponentialBackoffPolicy"/>.
+    /// </summary>
+    /// <exception cref="InvalidBackoffOptionsException">Thrown when one or more invalid options are found</exception>
+    internal void Validate()
+    {
+        var errorMessages = new List<string>();
+
+        if (CapBackoffTime < 0)
```
96malhar, line 25: Should this be changed to `CapBackoffTime <= 0`. Having a `CapBackoffTime` of 0 does not make sense.

src/AWS.Messaging/Services/Backoff/Policies/Options/IntervalBackoffOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Validates that the options set for the <see cref="IntervalBackoffPolicy"/>.
+    /// </summary>
+    /// <exception cref="InvalidBackoffOptionsException">Thrown when one or more invalid options are found</exception>
+    internal void Validate()
+    {
+        var errorMessages = new List<string>();
+
+        if (FixedInterval < 0)
```
96malhar, line 25: Should this be changed to `FixedInterval <= 0`. Having a `FixedInterval ` of 0 does not make sense.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: Even though I am checking for nulls on 249, Rider is still complaining on 251 that the value might be null which is why I added ??.

src/AWS.Messaging/Services/Backoff/BackoffHandler.cs
philasmar: Done

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
philasmar: Done

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
philasmar: I purposefully left in 0 as it is a valid value. A user could turn off backoffs by setting to 0.

src/AWS.Messaging/Services/Backoff/Policies/Options/IntervalBackoffOptions.cs
philasmar: I purposefully left in 0 as it is a valid value. A user could turn off backoffs by setting to 0.

96malhar (approved):

src/AWS.Messaging/Services/Backoff/IBackoffPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+/// </summary>
+public interface IBackoffPolicy
+{
+    /// <summary>
+    /// Determines if a <see cref="IBackoffHandler"/> should perform a backoff.
+    /// </summary>
+    /// <param name="exception">The exception that triggered a backoff in the <see cref="IBackoffHandler"/>.</param>
+    /// <param name="configuration">Internal configuration for polling messages from SQS.</param>
+    /// <returns>A boolean that indicates whether the <see cref="IBackoffHandler"/> should backoff or not.</returns>
+    bool ShouldBackoff(Exception exception, SQSMessagePollerConfiguration configuration);
```
normj, line 20: For the most part the backoff framework you have here is general purpose so we could use it for anywhere we want to do backoff. If we wanted to add another area in the project that to have backoff would we add a separate `ShouldBackoff` method?

src/AWS.Messaging/Services/Backoff/IBackoffHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+public interface IBackoffHandler
+{
+    /// <summary>
+    /// Performs a back off in the event that a given delegate throws an exception.
+    /// The delegate will be retried after backing off for a certain period of time.
+    /// </summary>
+    /// <param name="task">The delegate for which to perform a backoff.</param>
+    /// <param name="configuration">Internal configuration for polling messages from SQS.</param>
+    /// <param name="token">The cancellation token used to cancel the request.</param>
+    Task BackoffAsync(Func<Task> task, SQSMessagePollerConfiguration configuration, CancellationToken token);
```
normj, line 23: Should we have the `Func` return data? Something like the following would allow you to not have to predeclare the received message variable in the poller and then use it in the closure. Not sure if we would ever have a use case where we don't need to return someting.

```
Task<T> BackoffAsync<T>(Func<Task, T> task, SQSMessagePollerConfiguration configuration, CancellationToken token);
```

src/AWS.Messaging/Services/Backoff/IBackoffPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+
+    /// <summary>
+    /// Performs a calculation based on the number of retries to determine how long the <see cref="IBackoffHandler"/>
+    /// will wait before performing the next backoff.
+    /// </summary>
+    /// <param name="numberOfRetries">The number of times the <see cref="IBackoffHandler"/> has retried a request after performing a backoff.</param>
+    /// <returns>A number that indicates how long the <see cref="IBackoffHandler"/> should wait.</returns>
+    double RetrieveBackoffTime(int numberOfRetries);
```
normj, line 28: Return a `TimeSpan` instead of an ambiguous `double`.

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// <summary>
+    /// Retrieves the backoff time in seconds which is exponentially increasing based on the number of retries already performed.
+    /// </summary>
+    /// <param name="numberOfRetries">The number of times the <see cref="BackoffHandler"/> has retried a request after performing a backoff.</param>
+    /// <returns>An exponentially increasing backoff time in seconds.</returns>
+    public double RetrieveBackoffTime(int numberOfRetries)
+    {
+        var backoffTime = Math.Pow(2, numberOfRetries);
```
normj, line 47: Add some jitter to the backoffTime. Here is how the SDK does jitter.
https://github.com/aws/aws-sdk-net/blob/32f1a09b6f7b8e2f2e9b53332ad82e654927453e/sdk/src/Core/Amazon.Runtime/Pipeline/RetryHandler/StandardRetryPolicy.cs#L241

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+/// </summary>
+public class CappedExponentialBackoffOptions
+{
+    /// <summary>
+    /// The backoff time in seconds to cap the exponential backoff at.
+    /// Once the cap is reached, an interval backoff will occur with
+    /// <see cref="CapBackoffTime"/> as the fixed interval.
+    /// The default value is 1 hour.
+    /// </summary>
+    public int CapBackoffTime { get; set; } = 3600;
```
normj, line 17: An hour long backoff time is way too long. I would think more like 60 seconds as a default.

Also change the `int` to a `TimeSpan` so the units are not ambiguous. Unless that will make the deserialization from IConfiguration weird.

src/AWS.Messaging/Services/Backoff/Policies/Options/IntervalBackoffOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+/// <summary>
+/// Configuration for the interval backoff policy.
+/// </summary>
+public class IntervalBackoffOptions
+{
+    /// <summary>
+    /// The fixed interval in seconds to wait between back-offs.
+    /// The default value is 1 second.
+    /// </summary>
+    public int FixedInterval { get; set; } = 1;
```
normj, line 15: Same comment as above. Use `TimeSpan` unless that makes the deserialization from IConfiguration weird.

normj (changes requested):

src/AWS.Messaging/Services/Backoff/IBackoffPolicy.cs
philasmar: You can go about this in multiple ways. If your back off logic is dependent on the exception thrown, you would expand your implementation of the `ShouldBackoff` method to account for the different areas that a backoff is needed. However, if your logic is different, it makes sense to implement a new instance of `IBackoffPolicy` and pass it to a new `IBackoffHandler`. You could also do away with the `BackoffPolicy` and handle things differently in a new `IBackoffHandler`.

src/AWS.Messaging/Services/Backoff/IBackoffHandler.cs
philasmar: Fixed

src/AWS.Messaging/Services/Backoff/IBackoffPolicy.cs
philasmar: Fixed

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
philasmar: Fixed

src/AWS.Messaging/Services/Backoff/Policies/Options/CappedExponentialBackoffOptions.cs
philasmar: I updated the cap to 60 seconds. However I did not change the type to `TimeSpan` as it would make deserialization weird. This is kept consistent with the SQS Poller Configuration options which are all `int`.

src/AWS.Messaging/Services/Backoff/Policies/Options/IntervalBackoffOptions.cs
philasmar: I did not change the type to `TimeSpan` as it would make deserialization weird. This is kept consistent with the SQS Poller Configuration options which are all `int`.

normj: Ok

normj (approved):

src/AWS.Messaging/Services/Backoff/IBackoffHandler.cs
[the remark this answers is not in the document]
ashovlin: As long as there's some system 👍

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <summary>
+    /// Retrieves the backoff time in seconds which is exponentially increasing based on the number of retries already performed.
+    /// </summary>
+    /// <param name="numberOfRetries">The number of times the <see cref="BackoffHandler"/> has retried a request after performing a backoff.</param>
+    /// <returns>An exponentially increasing backoff time as a <see cref="TimeSpan"/>.</returns>
+    public TimeSpan RetrieveBackoffTime(int numberOfRetries)
+    {
+        double jitter;
+        lock (_randomJitter) {
```
ashovlin, line 49: Could we use `Random.Shared` to avoid needing to roll our own thread safety?

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,55 @@
...
+            jitter = _randomJitter.NextDouble();
+        }
+        var backoffTime = Convert.ToInt32(Math.Min(jitter * Math.Pow(2, numberOfRetries), _options.CapBackoffTime));
```
ashovlin, line 52: Curious about the source of this equation. It doesn't appear to be applying the jitter when we're at the `CapBackoffTime` - so those requests may become predicatable. Is that a concern?

The formula in https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ avoids this by applying the jitter to the outcome of `Math.Min`?

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,55 @@
...
+
+/// <summary>
+/// A backoff policy that will wait an exponentially increasing time for every backoff
+/// until it reaches a user-defined cap. At that point, it will switch to an interval backoff
+/// with the fixed interval equal to the cap.
+/// </summary>
+internal class CappedExponentialBackoffPolicy : IBackoffPolicy
+{
+    private readonly CappedExponentialBackoffOptions _options;
+    private readonly Random _randomJitter = new();
```
ashovlin, line 17: Will this be consistent per host?

From https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/:
> When adding jitter to scheduled work, we do not select the jitter on each host randomly. Instead, we use a consistent method that produces the same number every time on the same host.

src/AWS.Messaging/Services/Backoff/Policies/IntervalBackoffPolicy.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /// <summary>
+    /// Retrieves the backoff time in seconds which is a fixed interval, regardless of how many retries have already been performed.
+    /// </summary>
+    /// <param name="numberOfRetries">The number of times the <see cref="BackoffHandler"/> has retried a request after performing a backoff.</param>
+    /// <returns>A fixed interval representing the backoff time as a <see cref="TimeSpan"/>.</returns>
+    public TimeSpan RetrieveBackoffTime(int numberOfRetries)
+    {
+        return TimeSpan.FromSeconds(_options.FixedInterval);
```
ashovlin, line 45: Should we apply jitter here too? If running pollers using the interval policy on a fleet of hosts, is there a chance that that they could synchronize and retry polling at the same time?

>we found that clients frequently send requests on a regular interval, like once per minute. However, when a client has multiple servers behaving the same way, they can line up and trigger their requests at the same time. This can be the first few seconds of a minute, or the first few seconds after midnight for daily jobs. By paying attention to per-second load, and working with clients to jitter their periodic workloads, we accomplished the same amount of work with less server capacity.

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
philasmar: I am now applying jitter to `Math.Min`

src/AWS.Messaging/Services/Backoff/Policies/CappedExponentialBackoffPolicy.cs
philasmar: I don't believe so based on the implementation of `Random.Shared`. I will wait for customer feedback on this and see if we need to worry about it.

src/AWS.Messaging/Services/Backoff/Policies/IntervalBackoffPolicy.cs
philasmar: I applied jitter to the interval policy as well.

ashovlin (approved):

PR #112, opened 2024-03-22
ci: add support for automated versioning and changelog creation

philasmar:
*Description of changes:*
This PR adds support for AutoVer which automates versioning based on `change file` which contributors add to their PRs/Branches to indicate the type of change they are performing as well as the changelog message.
The change file looks like:
```
{
  "Projects": [
    {
      "Name": "AWS.Messaging",
      "Type": "Patch",
      "ChangelogMessages": [
        "The changelog message"
      ]
    }
  ]
}
```

As part of the normal SDLC on this repo, when we are ready to release a set of changes, we can kick off the `Create Release PR` workflow against the `dev` branch. This workflow will go through the `change file` and determine the proper version for the projects that need to be versioned. It will also create the changelog based on the messages in the `change files`. These file changes will be added to a new `releases/next-release` branch for which the workflow will create a PR against to be merged to `dev`. We will have a chance to review the release PR as a sanity check. Once we are satisfied with the versions and changelog, we can merge that PR into dev.
This will trigger the workflow `Sync dev and main` which will check if the Release PR was successfully merged or it was closed. If the PR was successfully merged, it will sync dev and main, create a GitHub release and delete the `releases/next-release` branch. If the PR was closed, it will clean up the tag and branch created by AutoVer.

AutoVer is configured using the file `.autover/autover.json`. In that file, I have specified the projects that we want to version and I have indicated the we will use `change files` to determine the version and changelogs.

### DIFF
--- a/.autover/autover.json
+++ b/.autover/autover.json
@@ -1,0 +1,19 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Path": "src/AWS.Messaging/AWS.Messaging.csproj"
+    },
+    {
+      "Name": "AWS.Messaging.Lambda",
+      "Path": "src/AWS.Messaging.Lambda/AWS.Messaging.Lambda.csproj"
+    },
+    {
+      "Name": "AWS.Messaging.Telemetry.OpenTelemetry",
+      "Path": "src/AWS.Messaging.Telemetry.OpenTelemetry/AWS.Messaging.Telemetry.OpenTelemetry.csproj"
+    }
+  ],
+  "UseCommitsForChangelog": false,
+  "DefaultIncrementType": "Patch",
+  "ChangeFilesDetermineIncrementType": true
+}
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -1,0 +1,71 @@
+# This GitHub Workflow will create a new release branch that contains the updated C# project versions and changelog.
+# The workflow will also create a PR that targets `dev` from the release branch.
+name: Create Release PR
+
+# This workflow is manually triggered when in preparation for a release. The workflow should be dispatched from the `dev` branch.
+on:
+  workflow_dispatch:
+
+jobs:
+  release-pr:
+    name: Release PR
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout a full clone of the repo
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: '0'
+      # Install .NET8 which is needed for AutoVer
+      - name: Setup .NET 8.0
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 8.0.x
+      # Install AutoVer to automate versioning and changelog creation
+      - name: Install AutoVer
+        run: dotnet tool install --global AutoVer --version 0.0.19
+      # Set up a git user to be able to run git commands later on
+      - name: Setup Git User
+        run: |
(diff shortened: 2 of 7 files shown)


.autover/autover.json
```diff
@@ -0,0 +1,19 @@
...
+      "Name": "AWS.Messaging.Lambda",
+      "Path": "src/AWS.Messaging.Lambda/AWS.Messaging.Lambda.csproj"
+    },
+    {
+      "Name": "AWS.Messaging.Telemetry.OpenTelemetry",
+      "Path": "src/AWS.Messaging.Telemetry.OpenTelemetry/AWS.Messaging.Telemetry.OpenTelemetry.csproj"
+    }
+  ],
+  "UseCommitsForChangelog": false,
+  "DefaultIncrementType": "Patch",
```
normj, line 17: Do we need a default if the change files are required to always say whether a patch or minor?

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,71 @@
...
+          git config --global user.name "aws-sdk-dotnet-automation"
+      # Create the release branch which will contain the version changes and updated changelog
+      - name: Create Release Branch
+        id: create-release-branch
+        run: |
+          branch=releases/next-release
+          git checkout -b $branch
+          echo "BRANCH=$branch" >> $GITHUB_OUTPUT
+      # Update the version of projects based on the change files
+      - name: Increment Version
```
normj, line 41: When do the change files get deleted? I would have thought it would have been part of this step but then the next section to update the changelog wouldn't have the change files to update the Changelog.

.autover/autover.json
philasmar: Creating change files using the command 'autover change' defaults to Patch for now. I can add a switch in the future to control that. The command however is meant to help users get started with creating a change file and they can still edit it manually. In the future, I plan to enrich this feature to have more options and allow users to do more using the CLI.

.github/workflows/create-release-pr.yml
philasmar: The change files get deleted when you run 'autover changelog'. `autover version` updates the project versions and adds a git tag. When you run 'autover changelog', the command looks for the git tag and checks the files in that commit. It reads the change files, creates the changelog and then deletes the files.

normj: If that is how the tool works now then I'm fine with how this workflow goes. For feedback on the `autover` tool that seems arbitrary that the change files are deleted as part of updating the changelog. I would think either the whole operation should be atomic so the `version` command would have an option to also update the change log or each operation is done separate to avoid surprises and deleting the change files would be a separate action.

philasmar: Thanks for the feedback. Initially when I started working on the tool, I had everything running under one command but it became a bit bloated, so to give users more control on what actions they want performed I separated the commands. I'll give the whole process another look to see how I can improve things, but this shouldn't have an impact on us moving forward with this PR.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: autover version
+        if: github.event.inputs.OVERRIDE_VERSION == ''
+      # Update the version of projects based on the override version
+      - name: Increment Version
+        run: autover version --use-version ${{ github.event.inputs.OVERRIDE_VERSION }}
```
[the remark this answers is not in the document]
philasmar: Fixed

ashovlin (approved):

Issue #97, opened 2024-02-16
EventBridge -> SQS. Failed to create a MessageEnvelop


careri:
Hi,

I've got a simple setup where I publish an event to EventBridge.
A simple rule that matces a source and forwards to a SQS.

I'm running commit #96 
OS: Windows 11

Publishing works as expected. Consuming fails, since the detail of the EventBridge envelope is an object and not a string:
```
Category: AWS.Messaging.Serialization.EnvelopeSerializer
Message: Failed to create a MessageEnvelope
Exception: System.InvalidOperationException: The requested operation requires an element of type \u0027String\u0027, but the target element has type \u0027Object\u0027.
   at System.Text.Json.ThrowHelper.ThrowJsonElementWrongTypeException(JsonTokenType expectedType, JsonTokenType actualType)
   at System.Text.Json.JsonDocument.GetString(Int32 index, JsonTokenType expectedType)
   at AWS.Messaging.Serialization.EnvelopeSerializer.GetMessageEnvelopeConfiguration(Message sqsMessage)
   at AWS.Messaging.Serialization.EnvelopeSerializer.ConvertToEnvelopeAsync(Message sqsMessage)
```

I've fixed the failing line. Will create a pull request.

careri:
PR:
https://github.com/awslabs/aws-dotnet-messaging/pull/98

96malhar:
This issue has been addressed via https://github.com/awslabs/aws-dotnet-messaging/pull/114.

PR #114, opened 2024-03-25
fix: Correctly deserialize outer envelope when processing messages relayed through EventBridge

96malhar:
*Issue*
* https://github.com/awslabs/aws-dotnet-messaging/issues/97
* DOTNET-7432

*Description of changes:* 
This is a sample SQS message when messages are forwarded by EventBridge
```
{
  "version": "0",
  "id": "961ac8a3-a4ee-0e17-d8d9-36c81ae29955",
  "detail-type": "AWS.Messaging.IntegrationTests.Models.ChatMessage",
  "source": "/aws/messaging",
  "account": "000000000",
  "time": "2024-03-25T20:39:11Z",
  "region": "us-west-2",
  "resources": [],
  "detail": {
    "id": "526d5466-5f51-42ce-a64b-8a220a7f1b71",
    "source": "/aws/messaging",
    "specversion": "1.0",
    "type": "AWS.Messaging.IntegrationTests.Models.ChatMessage",
    "time": "2024-03-25T20:39:10.8491344+00:00",
    "data": "{\"MessageDescription\":\"Test1\"}"
  }
}
```

The `detail` property is sent as a JSON object and not as a string. Invoking [JsonElement.GetString](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getstring?view=net-8.0) on this node throws an `InvalidOperationException`.

### DIFF
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -316,7 +316,15 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
     {
         if (node.TryGetProperty(propertyName, out var propertyValue))
         {
-            return propertyValue.ToString();
+            return propertyValue.ValueKind switch
+            {
+                JsonValueKind.Object => propertyValue.GetRawText(),
+                JsonValueKind.String => propertyValue.GetString(),
+                JsonValueKind.Number => propertyValue.ToString(),
+                JsonValueKind.True => propertyValue.ToString(),
+                JsonValueKind.False => propertyValue.ToString(),
+                _ => throw new InvalidDataException($"{propertyValue.ValueKind} cannot be converted to a string value"),
+            };
         }
         return null;
     }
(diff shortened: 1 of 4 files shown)


src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -316,7 +316,7 @@ private void SetEventBridgeMetadata(MessageEnvelopeConfiguration envelopeConfigu
     {
         if (node.TryGetProperty(propertyName, out var propertyValue))
         {
-            return propertyValue.GetString();
+            return propertyValue.ToString();
```
normj, line 319: Can we be more exact, I'm worried we could get some surprise from the `ToString()` method. Something like the following.

```
            if (propertyValue.ValueKind == JsonValueKind.String)
                return propertyValue.GetString();
            else if (propertyValue.ValueKind == JsonValueKind.Object)
                return propertyValue.GetRawText();
            else
                throw new Exception("Some invalid format with the right exception");
```

normj (changes requested):

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
96malhar: Done

normj (approved):

philasmar (approved):

ashovlin (approved):

PR #130, opened 2024-04-16
return the result from the publisher to the caller.

robvanpamel:
*Description of changes:*

When a message fails from eventbridge, the error result was not shown. This could lead to some issues in the future. 
I adapted this so the errormessage and if required the errorcode can be retrieved 

As the publisher is abstracted across EventBridge, SQS and SNS I had to make some sacrifices for naming; event <-> message. 
Please have a look at it and provide some feedback so it can be adapted where you need it. 

### DIFF
--- a/src/AWS.Messaging/Exceptions.cs
+++ b/src/AWS.Messaging/Exceptions.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using AWS.Messaging.Configuration;
+using AWS.Messaging.Publishers.EventBridge;
 using AWS.Messaging.Serialization;
 using Microsoft.Extensions.Configuration;
 
@@ -118,7 +119,7 @@ public class FailedToSerializeApplicationMessageException : AWSMessagingExceptio
 }
 
 /// <summary>
-/// Thrown if an exception occurs while publishing the message to SNS, SQS or EventBridge.
+/// Thrown if an exception occurs while publishing the message.
 /// </summary>
 public class FailedToPublishException : AWSMessagingException
 {
@@ -273,9 +274,9 @@ public class InvalidFifoPublishingRequestException : AWSMessagingException
 }
 
 /// <summary>
-/// Thrown if the EventBridgePublishResponse contains a message with an errorcode
+/// Thrown if the <see cref="EventBridgePublishResponse"/> contains a message with an error code
 /// </summary>
-public class EventBridgePutEventsException : Exception
+public class EventBridgePutEventsException : AWSMessagingException
 {
     /// <summary>
     /// The error code from the EventBridge SDK
--- a/src/AWS.Messaging/Exceptions.cs
+++ b/src/AWS.Messaging/Exceptions.cs
@@ -118,6 +118,17 @@ public class FailedToSerializeApplicationMessageException : AWSMessagingExceptio
 }
 
 /// <summary>
+/// Thrown if an exception occurs while publishing the message to SNS, SQS or EventBridge.
+/// </summary>
+public class FailedToPublishException : AWSMessagingException
+{
+    /// <summary>
+    /// Creates an instance of <see cref="FailedToPublishException"/>.
+    /// </summary>
+    public FailedToPublishException(string message, Exception? innerException = null) : base(message, innerException) { }
+}
+
+/// <summary>
 /// Thrown if the message being sent is invalid.
 /// </summary>
 public class InvalidMessageException : AWSMessagingException
@@ -260,3 +271,23 @@ public class InvalidFifoPublishingRequestException : AWSMessagingException
     /// </summary>
     public InvalidFifoPublishingRequestException(string message, Exception? innerException = null) : base(message, innerException) { }
 }
+
+/// <summary>
+/// Thrown if the EventBridgePublishResponse contains a message with an errorcode
+/// </summary>
+public class EventBridgePutEventsException : Exception
+{
+    /// <summary>
+    /// The error code from the EventBridge SDK
+    /// </summary>
+    public string ErrorCode { get; }
+
+    /// <summary>
+    /// Creates an instance of <see cref="EventBridgePutEventsException"/>.
+    /// </summary>
+    public EventBridgePutEventsException(string message, string errorCode) : base(message)
+    {
+        ErrorCode = errorCode;
+    }
+}
+
--- a/src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
+++ b/src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
@@ -4,8 +4,6 @@
 namespace AWS.Messaging.Publishers.EventBridge;
 /// <summary>
 /// Represents the results of an event published to an event bus.
-///
-///
 /// <para>
 /// If the publishing was successful, the entry has the event ID in it. Otherwise, an exception will be thrown
 /// </para>
@@ -17,9 +15,9 @@ namespace AWS.Messaging.Publishers.EventBridge;
 public class EventBridgePublishResponse : IPublishResponse
 {
     /// <summary>
-    /// Gets and sets the property EventId.
+    /// Gets and sets the property MessageId.
     /// <para>
-    /// The ID of the event.
+    /// The ID of the message.
     /// </para>
     /// </summary>
     public string? MessageId { get; set; }
--- a/src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
+++ b/src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
@@ -7,10 +7,8 @@ namespace AWS.Messaging.Publishers.EventBridge;
 ///
 ///
 /// <para>
-/// If the publishing was successful, the entry has the event ID in it. Otherwise, you
-/// can use the error code and error message to identify the problem with the entry.
+/// If the publishing was successful, the entry has the event ID in it. Otherwise, an exception will be thrown
 /// </para>
-///
 /// <para>
 /// For information about the errors that are common to all actions, see <a href="https://docs.aws.amazon.com/eventbridge/latest/APIReference/CommonErrors.html">Common
 /// Errors</a>.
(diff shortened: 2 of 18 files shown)


normj:
For SNS and SQS a user would need to catch exceptions and for EventBridge they would need to check the response which is likely to be forgotten. For example we forgot to think about it. I believe EventBridge's `PutEvents` API does the HTTP 200 with errors pattern because it supports batch event publishing. Since we are only sending a single event I wonder if we should have the EventBridge publisher check if response for failure and then we should throw a Messaging exception. Update the other publishers to also throw the same Messaging exception wrapping the SDK's exception. What do you think?

If we decide to support the batch flow in the future we can handle that threw separate APIs.

robvanpamel:
I agree to create a separate API for batch processing in the future. 
So to continue, we keep the existing `IPublishResponse ` to return a messageid / eventid but remove the ErrorMessage from it. I do think the messageid is required as a lot of people use this. At least on my 2 previous projects we had this in place. 

For returning the errormessage to the user with eventbridge, we check the response and throw a Message Exception. Let's call it `FailedToPublishException`. ( Like the others it would inherit from `AWSMessagingException`) 

For the other publishers SQS and SNS, we catch the exceptions and wrap it with the same `FailedToPublishException`.  
Do you mind if the existing exceptions, eg `InvalidMessageException` would be wrapped in the `FailedToPublishException` as an innerexception as well (https://github.com/awslabs/aws-dotnet-messaging/blob/main/src/AWS.Messaging/Publishers/SNS/SNSPublisher.cs )? That way it can be easily implemented.

normj:
I agree we should have the `IPublishResponse` with `MessageId`. I'm preferring `MessageId` over `EventId` since we are calling this our messaging library.

Yes to your logic for checking the response from event bridge and throw `FailedToPublishException` exception if there is a failure.

All publishers should throw `FailedToPublishException` and the XML docs for the publish method should indicate that with the `<exception>` tag. For SQS and SNS it will wrap the service exception. The `FailedToPublishException` exception when only be for for if the service call fails in some way. The `InvalidMessageException` is for bad configuration and I don't think should be wrapped in the `FailedToPublishException`.

robvanpamel:
I made the requested changes, let me know if there are things you would like to change

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
```diff
@@ -0,0 +1,129 @@
...
+        ///  <code>NotAuthorizedForDetailTypeException</code>
+        /// </para>
+        ///
+        /// <para>
+        /// You do not have permissions to publish events with this detail type onto this event
+        /// bus.
+        /// </para>
+        ///  </li> </ul>
+        /// </summary>
+    public string? ErrorCode { get; set; }
```
normj, line 112: Since an exception will always be thrown if there is an `ErrorCode` users will never get a response with `ErrorCode` and `ErrorMessage` set. I think we should remove them from the response to avoid confusion.

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
```diff
@@ -79,7 +82,7 @@ public async Task PublishAsync<T>(T message, EventBridgeOptions? eventBridgeOpti
                 if (string.IsNullOrEmpty(eventBusName))
                 {
                     _logger.LogError("Unable to determine a destination event bus for message of type '{MessageType}'.", typeof(T));
-                    throw new InvalidPublisherEndpointException($"Unable to determine a destination queue for message of type '{typeof(T)}'.");
+                    throw new InvalidPublisherEndpointException($"Unable to determine a destination event bus for message of type '{typeof(T)}'.");
```
normj, line 85: Good catch

src/AWS.Messaging/Publishers/SNS/SNSPublishResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+
+/// <summary>
+/// Response for Publish action.
+/// </summary>
+public class SNSPublishResponse : IPublishResponse
+{
+    /// <summary>
+    /// The error message as provided by SNS, if any
+    /// </summary>
+    public string? ErrorMessage { get; set; }
```
normj, line 14: I don't think this is ever set and can be removed.

src/AWS.Messaging/Publishers/SQS/SQSSendResponse.cs
```diff
@@ -0,0 +1,25 @@
...
+
+/// <summary>
+///
+/// </summary>
+public class SQSSendResponse : IPublishResponse
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public string? ErrorMessage { get; set; }
```
normj, line 14: Not set and can be removed.

normj (changes requested):

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
robvanpamel: yes, sorry I forgot to remove these !

robvanpamel:
Changes are done. I did had some issues in GIT so the local history looks strange. But I assume you can squash it, so that does not causes issues. 
Let me know if you see something else!

normj (approved):

normj:
Thanks @robvanpamel. Don't worry about the commits, we will squash when we merge. The PR looks good to me. Let me get another team member on the team to review it.

src/AWS.Messaging/Exceptions.cs
```diff
@@ -260,3 +271,23 @@ public class InvalidFifoPublishingRequestException : AWSMessagingException
     /// </summary>
     public InvalidFifoPublishingRequestException(string message, Exception? innerException = null) : base(message, innerException) { }
 }
+
+/// <summary>
+/// Thrown if the EventBridgePublishResponse contains a message with an errorcode
+/// </summary>
+public class EventBridgePutEventsException : Exception
```
ashovlin, line 278: Can this subclass `AWSMessagingException`? I think we use that to distinguish between framework-owned vs. "external" exceptions at some points.

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </para>
+/// <para>
+/// For information about the errors that are common to all actions, see <a href="https://docs.aws.amazon.com/eventbridge/latest/APIReference/CommonErrors.html">Common
+/// Errors</a>.
+/// </para>
+/// </summary>
+public class EventBridgePublishResponse : IPublishResponse
+{
+    /// <summary>
+    /// Gets and sets the property EventId.
```
ashovlin, line 20: Minor, could you update the doc comment to match `MessageId`?

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
```diff
@@ -110,12 +113,27 @@ public async Task PublishAsync<T>(T message, EventBridgeOptions? eventBridgeOpti
...
+                    return publishResponse;
+                }
+                _logger.LogDebug("The message of type '{MessageType}' has been pushed to EventBridge but failed with '{ErrorCode}'.", typeof(T), firstEntry.ErrorCode);
+                throw new EventBridgePutEventsException(firstEntry.ErrorMessage, firstEntry.ErrorCode);
             }
             catch (Exception ex)
             {
                 trace.AddException(ex);
+                if (ex is not AWSMessagingException)
+                    throw new FailedToPublishException("Message failed to publish.", ex);
```
ashovlin, line 136: Could this follow the same pattern as the other two, where `AmazonEventBridgeException` is the special case? 

If there's a reason why this is different (is this the 200 with errors case?), could you add that context as a comment?

src/AWS.Messaging/Exceptions.cs
```diff
@@ -260,3 +271,23 @@ public class InvalidFifoPublishingRequestException : AWSMessagingException
     /// </summary>
     public InvalidFifoPublishingRequestException(string message, Exception? innerException = null) : base(message, innerException) { }
 }
+
+/// <summary>
+/// Thrown if the EventBridgePublishResponse contains a message with an errorcode
```
philasmar, line 276: nit: use a doc reference for EventBridgePublishResponse
nit: separate "errorcode"

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublishResponse.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\r
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Messaging.Publishers.EventBridge;
+/// <summary>
+/// Represents the results of an event published to an event bus.
+///
+///
```
philasmar, line 8: nit: remove empty whitespace

philasmar:
Could you please rebase your PR on top of `dev`?
Also, could you please follow https://github.com/awslabs/aws-dotnet-messaging/blob/dev/CONTRIBUTING.md in order to add a `change file` in your PR?

src/AWS.Messaging/Publishers/EventBridge/EventBridgePublisher.cs
robvanpamel: I did this to make sure that other exception thrown by the AWS SDK which were not from the aws-dotnet-messaging framework were wrapped inside the FailedToPublishException. Since errors will show up in the error code it can be changed I assume. I don't know the internals of the EventBridge SDK very well. 

I changed it now as you requested.

robvanpamel:
I rebased, added the change file and pointed to dev iso main

philasmar (approved):

ashovlin (approved):

robvanpamel:
Is any action from my side required to integrate this into the next release?

normj:
@robvanpamel Your side is good. We just need schedule a release to include your change. Thanks for the PR.

normj:
Version `0.9.1` is out with this PR. Thanks again for the contribution.

Issue #150, opened 2024-07-24
ILambdaMessaging: Reset message visibility timeout when handler returns failed status


brendonparker:
### Describe the feature

Not sure if this a feature or a bug.

It appears there is some handling for resetting the visibility timeout when using the SQS Poller. However, the same feature doesn't exist when using the `ILambdaMessaging.ProcessLambdaEventWithBatchResponseAsync`. 

### Use Case

When the handler returns a Failed status, the message visibility should be reset, so that it can be immediately re-attempted and doesn't block that message group from processing the next message.

### Other Information

Using FIFO queue

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS.Messaging (or related) package versions

AWS.Messaging.Lambda 0.9.1

### Targeted .NET Platform

.NET 8

### Operating System and version

AmazonLinux/Lambda

brendonparker:
Actually, it looks like this behavior is already called out as a potential TODO here:

awslabs/aws-dotnet-messaging/docs/docs/design/message-visibility-timeout-handling.md

Having some configurability around this would be nice. 
I can probably work around it in the short term, by manually changing the visibility back to zero.

brendonparker:
This is my current work-around. Rolling my own handling of this, Would be nice to be handled by "the framework".

```csharp
public async Task<SQSBatchResponse> HandleAsync(SQSEvent evnt, ILambdaContext context)
{
    using var scope = Host.Services.CreateScope();
    var lambdaMessaging = scope.ServiceProvider.GetRequiredService<ILambdaMessaging>();
    var batchResponse = await lambdaMessaging.ProcessLambdaEventWithBatchResponseAsync(evnt, context);
    await ChangeMessageVisibilityForFailures(evnt, scope, batchResponse);
    return batchResponse;
}

private static async Task ChangeMessageVisibilityForFailures(SQSEvent evnt, IServiceScope scope, SQSBatchResponse batchResponse)
{
    var failureCount = batchResponse.BatchItemFailures.Count;
    if (failureCount == 0) return;

    using var sqs = scope.Resolve<IAmazonSQS>();
    var lookup = evnt.Records.ToDictionary(x => x.MessageId);

    if (failureCount == 1)
    {
        await sqs.ChangeMessageVisibilityAsync(new()
        {
            QueueUrl = AwsMessagingExtensions.MainQueueUrl,
[...]

PR #151, opened 2024-07-25
Add configurable visibility timeout default when using partial batch failures

brendonparker:
Issue #150

*Description of changes:*

Adds a configuration option to [LambdaMessagingOptions](https://github.com/awslabs/aws-dotnet-messaging/blob/main/src/AWS.Messaging.Lambda/LambdaMessagingOptions.cs) for a `VisibilityTimeout` value when there are partial batch failures. In that scenario, the failed messages will have their message visibility set to this default value.

Prior to this change, the message visibility would not be set/changed. Which, if you were using a FIFO queue, would mean that other messages in that message group would be blocked until the original message visibility timeout occurred, which could be a while.

After this change, the message visibility will be changed on failure.

### DIFF
--- a/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
+++ b/src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
@@ -76,7 +76,7 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
                 var sqsEvent = _configuration.SQSEvent;
 
                 trace.AddMetadata(TelemetryKeys.QueueUrl, _configuration.SubscriberEndpoint);
-                
+
                 if (sqsEvent is null || !sqsEvent.Records.Any())
                 {
                     return _sqsBatchResponse;
@@ -218,7 +218,7 @@ internal class DefaultLambdaMessageProcessor : ILambdaMessageProcessor, ISQSMess
         if(!messages.Any())
         {
             return;
-        }
+        }    
 
         var request = new DeleteMessageBatchRequest
         {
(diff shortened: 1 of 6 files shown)


brendonparker:
I made this only applicable when `UseBatchResponse` is `true`. I _think_ it could be changed to work regardless of that, if so desired.

src/AWS.Messaging.Lambda/Services/DefaultLambdaMessageProcessor.cs
```diff
@@ -256,6 +258,48 @@ public async Task DeleteMessagesAsync(IEnumerable<MessageEnvelope> messages, Can
         }
     }
 
+    /// <summary>
+    /// If configured for <see cref="AWS.Messaging.Lambda.Services.LambdaMessageProcessorConfiguration.VisibilityTimeoutForBatchFailures"/>
+    /// then change the visibility timeout on the failed items in the batch response.
+    /// </summary>
+    private async Task ResetVisibilityTimeoutForFailures()
+    {
+        if (_configuration.SQSEvent == null) return;
```
normj, line 267: nit: Style wise can you either use brackets or put the `return` on a separate indented line?

normj (approved):
I have a minor style nit but the change looks good. Thanks for the PR!

philasmar (approved):

philasmar:
@brendonparker I have ran the integration tests and they have passed for this PR. Once you address Norm's comments, I can squash your commits and release the changes.

brendonparker:
@philasmar done

philasmar:
This change was released in **AWS.Messaging.Lambda (0.10.0)**.

Thank you for your contribution @brendonparker!

PR #169, opened 2025-01-06
Update logic to make DefaultMessageManager fail on InvalidMessageHandlerSignatureException

ghost:
*Issue #, if available:* DOTNET-7824

*Description of changes:*

### Problem
The issue is that when an exception occurs during dependency injection, we are swallowing that error and not showing the error to the user when they invoke the function. Specifically the issue can be reproduced with something like this

```
public void ConfigureServices(IServiceCollection services)
{
    services.AddAWSService<IAmazonDynamoDB>();
    services.AddSingleton<IThingDoer>(x =>
    {
        var ddb = x.GetRequiredService<IAmazonDynamoDB>();
        return new ThingDoer(ddb, Environment.GetEnvironmentVariable("TABLE_NAME"));
    });
    services.AddSingleton<IDependentThing>(x =>
    {
        var thingDoer = x.GetRequiredService<IThingDoer>();
        throw new InvalidOperationException("Blah blah"); <------------------ replicating an error ocurring during DependencyInjection
        return new DependentThing(thingDoer);
    });
    services.AddAWSMessageBus(builder =>
    {
        builder.AddSQSPublisher<GreetingMessage>(Environment.GetEnvironmentVariable("QUEUE_URL"));
        builder.AddLambdaMessageProcessor();
        builder.AddMessageHandler<GreetingMessageHandler, GreetingMessage>();
    });
}
```

the user would run the messaging tool and see this as the output

```
{
  "batchItemFailures": [
    {
      "itemIdentifier": "19dd0b57-b21e-4ac1-bd88-01bbb068cb78"
    }
  ]
}
```
which does not provide and detail on what went wrong.

### Issue
This line
```
handler = scope.ServiceProvider.GetService(subscriberMapping.HandlerType);
```
was the one throwing an `InvalidOperationException` when an issue occurred during DI initialization.

### Fix
In order to fix this, i catch the exception when dependency injection fails, and then eventually re-throw it, which then displays on the UI as an error. 

The reason for changing `GetService` to `GetRequiredService` is because `GetRequiredService` throws an exception when it cannot get the service (GetService i think is throwing the error too in this specific scenario because its an exception occuring in the constructor of something during DI), which can be caught in one place. Technically, i could have also try/catched the `GetService` too, but then i would have had some duplicate for when I catch the exception and when `handler` is null, so I just use `GetRequiredService` to make things easier.

**Note**: This change also affects the other [scenarios ](https://github.com/awslabs/aws-dotnet-messaging/blob/0541a95355918657c2c0255889a6804b3145074c/src/AWS.Messaging/Services/HandlerInvoker.cs#L57) which use the `InvalidMessageHandlerSignatureException` error. However, i believe these errors should also be treated as fatal, since that is what is listed [here](https://github.com/awslabs/aws-dotnet-messaging/blob/0541a95355918657c2c0255889a6804b3145074c/src/AWS.Messaging/Configuration/SQSMessagePollerConfiguration.cs#L167)

image

### DIFF
--- a/src/AWS.Messaging/Services/HandlerInvoker.cs
+++ b/src/AWS.Messaging/Services/HandlerInvoker.cs
@@ -56,9 +56,12 @@ public class HandlerInvoker : IHandlerInvoker
 
                 using (var scope = _serviceProvider.CreateScope())
                 {
-                    var handler = scope.ServiceProvider.GetService(subscriberMapping.HandlerType);
-
-                    if (handler == null)
+                    object handler;
+                    try
+                    {
+                        handler = scope.ServiceProvider.GetRequiredService(subscriberMapping.HandlerType);
+                    }
+                    catch
                     {
                         _logger.LogError("Unable to resolve a handler for {HandlerType} while handling message ID {MessageEnvelopeId}.", subscriberMapping.HandlerType, messageEnvelope.Id);
                         throw new InvalidMessageHandlerSignatureException($"Unable to resolve a handler for {subscriberMapping.HandlerType} " +
--- a/src/AWS.Messaging/Services/HandlerInvoker.cs
+++ b/src/AWS.Messaging/Services/HandlerInvoker.cs
@@ -61,11 +61,11 @@ public class HandlerInvoker : IHandlerInvoker
                     {
                         handler = scope.ServiceProvider.GetRequiredService(subscriberMapping.HandlerType);
                     }
-                    catch
+                    catch (Exception e)
                     {
                         _logger.LogError("Unable to resolve a handler for {HandlerType} while handling message ID {MessageEnvelopeId}.", subscriberMapping.HandlerType, messageEnvelope.Id);
                         throw new InvalidMessageHandlerSignatureException($"Unable to resolve a handler for {subscriberMapping.HandlerType} " +
-                                                                          $"while handling message ID {messageEnvelope.Id}.");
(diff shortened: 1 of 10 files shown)


src/AWS.Messaging/Services/HandlerInvoker.cs
```diff
@@ -56,9 +56,12 @@ public async Task<MessageProcessStatus> InvokeAsync(MessageEnvelope messageEnvel
...
                 {
-                    var handler = scope.ServiceProvider.GetService(subscriberMapping.HandlerType);
-
-                    if (handler == null)
+                    object handler;
+                    try
+                    {
+                        handler = scope.ServiceProvider.GetRequiredService(subscriberMapping.HandlerType);
+                    }
+                    catch
```
normj, line 64: We should include this exception as the inner exception for the `InvalidMessageHandlerSignatureException`.

normj (changes requested):

src/AWS.Messaging/Services/HandlerInvoker.cs
gcbeattyAWS: i've updated it to include the inner exception and have re-tested and verified that the error shows the inner exception as well in the ui. I have updated the PR description with the new screenshot too

normj (approved):

philasmar (approved):

Issue #174, opened 2025-02-14
Incorrect Handling of Unprocessed Messages with SQS FIFO Queue


max-gudkov:
### Describe the bug

We are encountering an issue with our AWS setup where an SQS FIFO queue is configured as a trigger for an AWS Lambda function. The trigger has batching enabled with a maximum of 10 messages and the "Report batch item failures" feature is also enabled. According to [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/services-sqs-errorhandling.html#services-sqs-batchfailurereporting): 

"If you're using this feature with a FIFO queue, your function should stop processing messages after the first failure and return all failed and unprocessed messages in batchItemFailures. This helps preserve the ordering of messages in your queue."

However, the current implementation behaves differently. The DefaultMessageManager stops processing messages withing the message group after the first failure as expected, but it only returns the failed messages in batchItemFailures. The unprocessed messages are treated by AWS SQS as successfully processed and are subsequently deleted from the queue. This behavior leads to loss of messages and potential data inconsistency.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The Lambda function should return both failed and unprocessed messages in batchItemFailures, and unprocessed messages should not be deleted from the queue.

### Current Behavior

Only failed messages are returned in batchItemFailures, and unprocessed messages are treated as successfully processed and deleted from the queue.

### Reproduction Steps

1. Configure an AWS SQS FIFO queue as a trigger for an AWS Lambda function.
2. Enable batching with and enable "Report batch item failures" when configure the Lambda function trigger.
3. Send multiple messages to the SQS FIFO queue with the same MessageGroupId.
4. Introduce a failure in processing one of the messages.
5. Observe that only failed messages are returned in batchItemFailures and unprocessed messages are deleted from the queue.

### AWS.Messaging (or related) package versions

AWS.Messaging.Lambda 0.10.0
AWS.Messaging 0.9.3

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11

ashishdhingra:
@max-gudkov Good afternoon. Thanks for opening the issue. While I'm trying to reproduce the issue, to ensure that we both are following the same steps, could you please share:
- Self-contained reproducible code.
- Exact sequence of steps along with message structure used to send SQS messages.

Thanks,
Ashish

normj:
[...]

PR #175, opened 2025-02-15
Fix issue with fifo queues message failure reporint issue

normj:
*Issue #, if available:*
https://github.com/awslabs/aws-dotnet-messaging/issues/174

*Description of changes:*
While processing messages in Lambda if we get a failure in a message group we skip processing the remaining messages. We should be reporting to Lambda as part of the SQSBatchResponse the items we skipped as failures so Lambda doesn't think they were successfully processed.

### DIFF
--- a/src/AWS.Messaging/Services/DefaultMessageManager.cs
+++ b/src/AWS.Messaging/Services/DefaultMessageManager.cs
@@ -133,7 +133,7 @@ public class DefaultMessageManager : IMessageManager
             var remaining = messageGroup.Count;
 
             // Sequentially process each message within the group
-            for(var i = 0; i < messageGroup.Count; i += 1)
+            for(var i = 0; i < messageGroup.Count; i++)
             {
                 var message = messageGroup[i];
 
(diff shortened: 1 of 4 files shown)


src/AWS.Messaging/Services/DefaultMessageManager.cs
```diff
@@ -133,12 +133,19 @@ public async Task ProcessMessageGroupAsync(List<ConvertToEnvelopeResult> message
             var remaining = messageGroup.Count;
 
             // Sequentially process each message within the group
-            foreach (var message in messageGroup)
+            for(var i = 0; i < messageGroup.Count; i += 1)
```
ashishdhingra, line 136: Just curious, if this logic is common for standard queues?

normj: Standard queues don't have message queues so don't come in to this method.

ashishdhingra (approved):

src/AWS.Messaging/Services/DefaultMessageManager.cs
boblodgett, line 136: Non blocking: `i++` is more common for incrementing by 1.

boblodgett (approved):

src/AWS.Messaging/Services/DefaultMessageManager.cs
normj: Agreed, not sure how that happened. I blame the intellisense :). Fixed now.

boblodgett (approved):

Issue #144, opened 2024-06-03
Type parameters should be decorated with DynamicallyAccessedMembersAttribute


NoahStolk:
### Describe the feature

The `MessageBusBuilder` class exposes various methods that accept `Type` instances or type parameters that are used to dynamically construct instances of said types. For example, in the `MessageBusBuilder.AddMessageHandler` method, the `THandler` type is used to dynamically construct an instance of that type.

This can and should be indicated by adding [DynamicallyAccessedMembersAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-8.0) to the type parameter. This allows tools such as ReSharper to detect that this type is created dynamically. Adding this attribute would also prevent code from being trimmed incorrectly when compiling to native code using Native AOT.

### Use Case

Currently, ReSharper complains that my handler type is never constructed and should be removed because it is seen as "dead code". I also believe (but have not yet tested this) that my handler type would be removed if I enabled trimming or compiled my application to native code using Native AOT, which would cause runtime exceptions. There are various workarounds for this, but the obvious fix would be to just include the attribute here.

### Proposed Solution

Add the attribute where necessary. Example usages of the attribute can be found in the ASP.NET Core runtime. For example, `AddSingleton`, `AddScoped`, and `AddTransient` use this attribute:

```cs
public static IServiceCollection AddSingleton<TService, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] TImplementation>(this IServiceCollection services)
    where TService : class
    where TImplementation : class, TService
// ...

public static IServiceCollection AddSingleton(
    this IServiceCollection services,
    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type serviceType)
// ...
```

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS.Messaging (or related) package versions

AWS.Messaging 0.9.1

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11

ashishdhingra:
Needs review with the team.

ashishdhingra:
@NoahStolk Thanks for the issue report. We have a larger effort in our internal backlog to make this library NativeAOT compatible. This is one the the things we would consider.

normj:
[...]

PR #184, opened 2025-03-07
Add support for Native AOT (Retarget library to .NET 8)

normj:
*Issue*
https://github.com/awslabs/aws-dotnet-messaging/discussions/109
https://github.com/awslabs/aws-dotnet-messaging/issues/144

*Description of changes:*
This PR enables `IsTrimmable` in the 3 packages and address all trim warnings. Note this is targeting the branch that will use V4 of the AWS SDK which has more Native AOT work done for that is required for this library.

The main difference from user's point of view is to use Native AOT they must use the `AddMessageBus` overload that takes in a `JsonSerializerContext`. That context must have the source generation setup for all of the POCOs used by the user that represent messages. If they do not use this overload the user will get a trim warning about the code requiring unreferenced code.

I also the libraries to target .NET 8.0 to get the SDK's Native AOT support. Since .NET 6.0 is out of support and we haven't GA the library yet we don't need to start the GA lifecycle on unsupported targets.

The loading messaging config from `IConfiguration` via the `LoadConfigurationFromSettings` method is not supported for trimming because that will always requiring loading types that are not known at compile time.

The trickiest part for adding trimming support was the use of `MakeGenericType` in EnvelopeSerializer. In that class we are taking the user's Type for the message and turning that into `MessageEnvelope<User'sType>`. That was using the type of reflection via the `MakeGenericType` and `Activator.CreateInstance` that doesn't work in a trimming environment. That is because when working with the `Type` object you can't know at compile time what the type is. To handle that I created an envelope factory on the `SubscriberMapping` that uses the generic parameters to create the `MessageEnvelope<User'sType>`. Since the generic parameters are known at compile time we don't have to use the `MakeGenericType` and `Activator.CreateInstance` methods.

## Testing
Ran unit and integ tests and updated tests to handle new code paths when appropiate.
Build a console application with `PublishAot` set to true. The code handled publishing and subscribe and no trim warnings were produced and application ran successful.

Code of console application
```csharp
using AWS.Messaging;
using MessagingLocalAOTTest;

var builder = new HostApplicationBuilder();

var queueUrl = "<queue-url>";
builder.Services.AddAWSMessageBus(MessageFunctionJsonSerializerContext.Default, builder =>
{
    builder.AddSQSPublisher<ProductDTO>(queueUrl);
    builder.AddMessageHandler<ProductHandler, ProductDTO>();

    builder.AddSQSPoller(queueUrl);
});

var app = builder.Build();

var task = app.RunAsync();


var publisher = app.Services.GetRequiredService<IMessagePublisher>();

var product = new ProductDTO("1", "Foo", "Bar");

await publisher.PublishAsync(product);

var cancelSource = new CancellationTokenSource();
cancelSource.CancelAfter(5000);

try
{
    Task.WaitAll(new Task[] { task }, cancelSource.Token);
    Console.WriteLine("Application completed");
}
catch (OperationCanceledException) { }
```

Json serializer context:
```csharp
using System.Text.Json.Serialization;

namespace MessagingLocalAOTTest;

public record ProductDTO(string ID, string Name, string Description);

[JsonSerializable(typeof(ProductDTO))]
[JsonSourceGenerationOptions(UseStringEnumConverter = true)]
public partial class MessageFunctionJsonSerializerContext : JsonSerializerContext
{
}

```

Message Handler:
```csharp
using AWS.Messaging;

namespace MessagingLocalAOTTest;

public class ProductHandler : IMessageHandler<ProductDTO>
{
    public Task<MessageProcessStatus> HandleAsync(MessageEnvelope<ProductDTO> messageEnvelope, CancellationToken token = default)
    {
        Console.WriteLine($"Saving product: {messageEnvelope.Message.Name}");
    }
}
```


src/AWS.Messaging/Services/IMessageManagerFactory.cs
```diff
@@ -40,6 +41,13 @@ public DefaultMessageManagerFactory(IServiceProvider serviceProvider)
     /// <inheritdoc/>
     public IMessageManager CreateMessageManager(ISQSMessageCommunication sqsMessageCommunication, MessageManagerConfiguration configuration)
     {
-        return ActivatorUtilities.CreateInstance<DefaultMessageManager>(_serviceProvider, sqsMessageCommunication, configuration);
+        var manager = new DefaultMessageManager(
```
normj, line 44: The `ActivatorUtilities.CreateInstance` call here actually compiled with out trim warnings but failed at the runtime. Easy fix was to directly create the instance instead of relying on `ActivatorUtilities.CreateInstance`.

src/AWS.Messaging/AWS.Messaging.csproj
```diff
@@ -15,20 +15,22 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <EnforceCodeStyleInBuild>True</EnforceCodeStyleInBuild>
     <WarningsAsErrors>CA1727</WarningsAsErrors>
+    <NoWarn>CS1591</NoWarn>
```
normj, line 18: CS1591 is the code for no docs on public API surface. I had to add this because the code generated via the JSON serializer for the internal `MessageJsonSerializerContext` doesn't have any docs. I tried a variety of ways to get the warning to ignore that generated code but nothing stuck.

test/AWS.Messaging.IntegrationTests/AWS.Messaging.IntegrationTests.csproj
```diff
@@ -1,24 +1,24 @@
...
-    <PackageReference Include="AWSSDK.S3" Version="4.0.0-preview.4" />
-    <PackageReference Include="AWSSDK.SecurityToken" Version="4.0.0-preview.4" />
+    <PackageReference Include="AWSSDK.CloudWatchLogs" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.Lambda" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.S3" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="4.0.0-preview.10" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.*" />
```
philasmar, line 19: nit: could you use .NET 8 versions?

src/AWS.Messaging/AWS.Messaging.csproj
```diff
@@ -15,20 +15,22 @@
...
-    <PackageReference Include="AWSSDK.Extensions.NETCore.Setup" Version="4.0.0-preview.4" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="4.0.0-preview.4" />
-    <PackageReference Include="AWSSDK.SQS" Version="4.0.0-preview.4" />
+    <PackageReference Include="AWSSDK.EventBridge" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.Extensions.NETCore.Setup" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.SQS" Version="4.0.0-preview.10" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="6.0.*" />
```
philasmar, line 36: could you use .NET 8 versions?

test/AWS.Messaging.Tests.Common/AWS.Messaging.Tests.Common.csproj
```diff
@@ -1,20 +1,20 @@
...
-    <PackageReference Include="AWSSDK.SQS" Version="4.0.0-preview.4" />
+    <PackageReference Include="AWSSDK.CloudWatchLogs" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.IdentityManagement" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.Lambda" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.S3" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.SecurityToken" Version="4.0.0-preview.10" />
+    <PackageReference Include="AWSSDK.SQS" Version="4.0.0-preview.10" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.*" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.*" />
```
philasmar, line 20: nit: same comment on .NET versions

philasmar (commented):
You are writing out types in the documentation instead of doing doc references. This is happening in multiple places. Could update those?

test/AWS.Messaging.IntegrationTests/AWS.Messaging.IntegrationTests.csproj
normj: Fixed

src/AWS.Messaging/AWS.Messaging.csproj
normj: Fixed

test/AWS.Messaging.Tests.Common/AWS.Messaging.Tests.Common.csproj
normj: Fixed

philasmar (approved):

GarrettBeatty (approved):

PR #189, opened 2025-03-14
Refactor code for EnvelopeSerializer. Remove JSONSerialize.Deserialize logic for whole message

GarrettBeatty:
*Issue #, if available:* DOTNET-7873

*Description of changes:* 
The main change is that we no longer use `JsonSerializer.Deserialize` on the whole `sqs.Body`. This was parsing both the envelope (Which contains metadata fields and things) and also the `data` field inside. Previously this was all fine because were always storing `data` as an encoded json string. However, since I am working on [changing ](https://github.com/awslabs/aws-dotnet-messaging/pull/190) that behavior in the future to store `data` as the actual json object, we can no longer use `JsonSerializer.Deserialize` on the whole body (because JsonSerializer.Deserialize fails). Instead we have to do:
1. Parse outer wrapper using JsonDocument.Parse. This will parse everything except the `data` field.
2. Deserialize `data` using `_messageSerializer.Deserialize`.

This way it allows users to implement and kind of message serializer logic for the `data` field. It could be json, xml, etc

I also made some improvements by creating separate class for each parser and metadata creator.

## Testing
1. Unit and integration tests pass
2. I made a console app to verify it still works

```

namespace AWS.Messaging.SerializationIssue;

internal class Program
{
    private static async Task Main(string[] args)
    {
        var builder = Host.CreateApplicationBuilder(args);

        builder.Services.AddAWSMessageBus(bus =>
        {
            bus.AddEventBridgePublisher<Payload>("arn:aws:events:us-east-1:147997163238:event-bus/new-bus", nameof(Payload));
            
            bus.AddSQSPublisher<Payload>("https://sqs.us-east-1.amazonaws.com/147997163238/newqueue");

            
            // Register an SQS Queue that the framework will poll for messages
            bus.AddSQSPoller("https://sqs.us-east-1.amazonaws.com/147997163238/newqueue");

            // Register all IMessageHandler implementations with the message type they should process. 
            // Here messages that match our ChatMessage .NET type will be handled by our ChatMessageHandler
            bus.AddMessageHandler<ChatMessageHandler, Payload>();
            
        });
        
      

        builder.Services.AddHostedService<Worker>();

        var host = builder.Build();
        await host.RunAsync();
    }
}

public class ChatMessageHandler : IMessageHandler<Payload>
{
    public Task<MessageProcessStatus> HandleAsync(MessageEnvelope<Payload> messageEnvelope, CancellationToken token = default)
    {
        // Add business and validation logic here
        if (messageEnvelope == null)
        {
            return Task.FromResult(MessageProcessStatus.Failed());
        }

        if (messageEnvelope.Message == null)
        {
            return Task.FromResult(MessageProcessStatus.Failed());
        }

        Payload message = messageEnvelope.Message;

        Console.WriteLine($"Message Description: {message.Id}");

        // Return success so the framework will delete the message from the queue
    }
}
```

```
using Microsoft.Extensions.Logging;

namespace AWS.Messaging.SerializationIssue;

internal class Worker(
    ILogger<Worker> logger,
    IHostApplicationLifetime lifetime,
    IMessagePublisher messagePublisher
)
    : IHostedService
{
    public async Task StartAsync(CancellationToken cancellationToken)
    {
        try
        {
            var payload = new Payload
            {
                Id = Guid.NewGuid(),
                Name = "Test",
                SomeValue = 42
            };
            await messagePublisher.PublishAsync(payload, cancellationToken);
        }
        catch (Exception e)
        {
            logger.LogError(e, "An error occured");
        }
        finally
        {
            lifetime.StopApplication();
        }
    }

    public Task StopAsync(CancellationToken cancellationToken)
    {
        return Task.CompletedTask;
    }
}
```

### DIFF
--- a/.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
+++ b/.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Refactor EnvelopeSerializerCode."
+      ]
+    }
+  ]
+}
--- a/.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
+++ b/.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
@@ -4,7 +4,7 @@
       "Name": "AWS.Messaging",
       "Type": "Patch",
       "ChangelogMessages": [
-        "Refactor EnvelopeSerializerCode."
+        "Refactor logic for serialization."
       ]
     }
   ]
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -6,6 +6,7 @@ using System.Text.Json;
 using System.Text.Json.Nodes;
 using Amazon.SQS.Model;
 using AWS.Messaging.Configuration;
+using AWS.Messaging.Serialization.Helpers;
 using AWS.Messaging.Services;
 using Microsoft.Extensions.Logging;
 using AWS.Messaging.Serialization.Parsers;
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -8,6 +8,7 @@ using Amazon.SQS.Model;
 using AWS.Messaging.Configuration;
 using AWS.Messaging.Services;
 using Microsoft.Extensions.Logging;
+using AWS.Messaging.Serialization.Parsers;
 
 namespace AWS.Messaging.Serialization;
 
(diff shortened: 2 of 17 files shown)


test/AWS.Messaging.UnitTests/SerializationTests/EnvelopeSerializerTests.cs
```diff
@@ -478,92 +479,127 @@ public async Task SerializeAsync_DataMessageLogging_NoError(bool dataMessageLogg
     [InlineData(true)]
     public async Task SerializeAsync_DataMessageLogging_WithError(bool dataMessageLogging)
```
GarrettBeatty, line 480: this test had to be changed because we no longer use jsonserialize.deserialize

.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Refactor logic for serialization."
```
philasmar, line 7: add more information on the actual changes. This change warrants a Minor increment.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -193,164 +168,110 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
         }
     }
 
-    private void ValidateMessageEnvelope<T>(MessageEnvelope<T>? messageEnvelope)
+    private MessageEnvelope DeserializeEnvelope(string envelopeString, Type messageType)
     {
-        if (messageEnvelope is null)
+        using var document = JsonDocument.Parse(envelopeString);
+        var root = document.RootElement;
+        var envelopeType = typeof(MessageEnvelope<>).MakeGenericType(messageType);
```
philasmar, line 175: This logic is not trim safe. It won't be compatible with the NativeAOT work that Norm is doing. We should sync up on this.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -193,164 +168,110 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
...
+            _logger.LogError(ex, "Failed to deserialize or validate MessageEnvelope");
+            throw new InvalidDataException("MessageEnvelope instance is not valid", ex);
         }
-
-        SetSQSMetadata(envelopeConfiguration, sqsMessage);
-        return envelopeConfiguration;
     }
 
-    private void SetSQSMetadata(MessageEnvelopeConfiguration envelopeConfiguration, Message sqsMessage)
+    private static string GetMessageTypeFromEnvelope(string json)
```
philasmar, line 207: nit: consider making this function more reusable by passing the propertyName as a parameter.

philasmar (changes requested):

.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
GarrettBeatty: updated

PR #192, opened 2025-03-19
Refactor serialization logic and remove JsonSerializer.Deserialize use for entire envelope

GarrettBeatty:
*Issue #, if available:* DOTNET-7873

*Description of changes:* 
The main change is that we no longer use `JsonSerializer.Deserialize` on the whole `sqs.Body`. This was parsing both the envelope (Which contains metadata fields and things) and also the `data` field inside. Previously this was all fine because were always storing `data` as an encoded json string. However, since I am working on changing that behavior in the [future ](https://github.com/awslabs/aws-dotnet-messaging/pull/193/files)to store `data` as the actual json object, we can no longer use `JsonSerializer.Deserialize` on the whole body (because JsonSerializer.Deserialize fails). Instead we have to do:
1. Parse outer wrapper using JsonDocument.Parse. This will parse everything except the `data` field.
2. Deserialize `data` using `_messageSerializer.Deserialize`.

This way it allows users to implement and kind of message serializer logic for the `data` field. It could be json, xml, etc

Other changes
1. Moved parsing logic into its own class (e.g. `SQSParser`, `SNSParser`, `EventBridgeParser`. 

I have also validated that there are no trim warnings and i created the below console app to to verify it works. 

## Testing
1. Unit and integration tests pass
2. I made a console app to verify it still works

```
using AWS.Messaging;
using MessagingLocalAOTTest;

var builder = new HostApplicationBuilder();

var queueUrl = "url";
builder.Services.AddAWSMessageBus(MessageFunctionJsonSerializerContext.Default,builder =>
{
    builder.AddSQSPublisher<ProductDTO>(queueUrl);
    builder.AddMessageHandler<ProductHandler, ProductDTO>();

    builder.AddSQSPoller(queueUrl);
});

var app = builder.Build();

var task = app.RunAsync();


var publisher = app.Services.GetRequiredService<IMessagePublisher>();

var product = new ProductDTO("1", "Foo", "Bar");

await publisher.PublishAsync(product);

var cancelSource = new CancellationTokenSource();
cancelSource.CancelAfter(5000);

try
{
    Task.WaitAll(new Task[] { task }, cancelSource.Token);
    Console.WriteLine("Application completed");
}
catch (OperationCanceledException) { }

```

```
using System.Text.Json.Serialization;

namespace MessagingLocalAOTTest;

public record ProductDTO(string ID, string Name, string Description);

[JsonSerializable(typeof(ProductDTO))]
[JsonSourceGenerationOptions(UseStringEnumConverter = true)]
public partial class MessageFunctionJsonSerializerContext : JsonSerializerContext
{
}
```
```
using AWS.Messaging;

namespace MessagingLocalAOTTest;

public class ProductHandler : IMessageHandler<ProductDTO>
{
    public Task<MessageProcessStatus> HandleAsync(MessageEnvelope<ProductDTO> messageEnvelope, CancellationToken token = default)
    {
        Console.WriteLine($"Saving product: {messageEnvelope.Message.Name}");
    }
}
```

```
PS C:\Users\gcbeatty\RiderProjects\ConsoleApp1> dotnet publish -r win-x64 -c Release
Restore complete (0.9s)
  ConsoleApp1 succeeded (21.6s) → ConsoleApp1\bin\Release\net9.0\win-x64\publish\

Build succeeded in 22.9s
```
```

PS C:\Users\gcbeatty\RiderProjects\ConsoleApp1> .\ConsoleApp1\bin\Release\net9.0\win-x64\publish\ConsoleApp1.exe
info: AWSSDK[0]   
      Found AWS options in IConfiguration
info: AWSSDK[0]
      Found credentials using the AWS SDK's default credential search
info: AWS.Messaging.Services.MessagePumpService[0]
      Starting polling: https://sqs.us-east-1.amazonaws.com/147997163238/newqueue
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Production
info: Microsoft.Hosting.Lifetime[0]
      Content root path: C:\Users\gcbeatty\RiderProjects\ConsoleApp1
Saving product: Foo

```


.github/workflows/doc-site.yml
```diff
@@ -17,31 +17,33 @@ permissions:
 concurrency:
   group: "pages"
   cancel-in-progress: false
-  
+
```
GarrettBeatty, line 20: this is just merging dev to this branch

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -183,169 +169,123 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
...
-            Source = GetJsonPropertyAsString(root, "source"),
-            DetailType = GetJsonPropertyAsString(root, "detail-type"),
-            Time = GetJsonPropertyAsDateTimeOffset(root, "time"),
-            AWSAccount = GetJsonPropertyAsString(root, "account"),
-            AWSRegion = GetJsonPropertyAsString(root, "region"),
-            Resources = GetJsonPropertyAsList<string>(root, "resources")
+            rootCopy = document.RootElement.Clone();
+        }
+
+        var parsers = new IMessageParser[]
```
normj, line 238: Instead of recreating this array of stateless parsers for every message should we make the array a static field of the `EnvelopeSerializer`.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -138,38 +139,23 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
...
-            if (subscriberMapping is null)
-            {
-                _logger.LogError("{MessageConfiguration} does not have a valid subscriber mapping for message ID '{MessageTypeIdentifier}'", nameof(_messageConfiguration), messageTypeIdentifier);
-                throw new InvalidDataException($"{nameof(_messageConfiguration)} does not have a valid subscriber mapping for {nameof(messageTypeIdentifier)} '{messageTypeIdentifier}'");
-            }
+            // Get the raw envelope JSON and metadata from the appropriate wrapper (SNS/EventBridge/SQS)
+            var (envelopeJson, metadata) = await ParseOuterWrapper(sqsMessage);
+
+            // Parse just the type field first to get the correct mapping
+            var messageType = GetMessageTypeFromEnvelope(envelopeJson);
```
normj, line 146: `GetMessageTypeFromEnvelope` is doing a full JSON document parse to just get the `type` and then we look up the mapping. Can't we put this logic in the `DeserializeEnvelope` which also does a JSON parse. We wouldn't be passing in the message type and mapping into `DeserializeEnvelope` it would determine those values after it does a parse.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -183,169 +169,123 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
...
-            TopicArn = GetJsonPropertyAsString(root, "TopicArn"),
-            Subject = GetJsonPropertyAsString(root, "Subject"),
-            UnsubscribeURL = GetJsonPropertyAsString(root, "UnsubscribeURL"),
-            Timestamp = GetJsonPropertyAsDateTimeOffset(root, "Timestamp")
-        };
-        if (root.TryGetProperty("MessageAttributes", out var messageAttributes))
-        {
-            envelopeConfiguration.SNSMetadata.MessageAttributes = messageAttributes.Deserialize(MessagingJsonSerializerContext.Default.DictionarySNSMessageAttributeValue);
-        }
+        using var doc = JsonDocument.Parse(json);
```
normj, line 223: See my comment above how we should avoid doing an additional parse of the document. I think we can get rid of this method and collapse this logic inside the deserialize envelope method.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -183,169 +169,123 @@ public async ValueTask<ConvertToEnvelopeResult> ConvertToEnvelopeAsync(Message s
...
+        using (var document = JsonDocument.Parse(sqsMessage.Body))
         {
-            EventId = GetJsonPropertyAsString(root, "id"),
-            Source = GetJsonPropertyAsString(root, "source"),
-            DetailType = GetJsonPropertyAsString(root, "detail-type"),
-            Time = GetJsonPropertyAsDateTimeOffset(root, "time"),
-            AWSAccount = GetJsonPropertyAsString(root, "account"),
-            AWSRegion = GetJsonPropertyAsString(root, "region"),
-            Resources = GetJsonPropertyAsList<string>(root, "resources")
+            rootCopy = document.RootElement.Clone();
```
normj, line 235: Why do we need to clone the element?

normj (changes requested):

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: i used clone because the variable was out of scope because i was doing `using` when parsing. But i guess i can just remove the `using` part so it doesnt get disposed

GarrettBeatty: for now i kept it as-is since i would need to update everywhere to use node.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: Ive updated the code to remove GetMessageTypeFromEnvelope  and just figure out the type in the same method as deserializeenvelope.

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: fixed

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: fixed

philasmar (approved):

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
normj: The messages could be quite large so cloning the whole message is something we really shouldn't do. I understand you need to dispose the json document but you don't need to dispose of till the end of the method. What about something like this:

```csharp
var document = JsonDocument.Parse(sqsMessage.Body);
try
{
    string currentMessageBody = sqsMessage.Body;
    var combinedMetadata = new MessageMetadata();

    // Try each parser in order
    foreach (var parser in _parsers.Where(p => p.CanParse(document.RootElement)))
    {
        // Example 1 (SNS message) flow:
        // 1. SNSMessageParser.CanParse = true (finds "Type": "Notification")
        // 2. parser.Parse extracts inner message and SNS metadata
        // 3. messageBody = contents of "Message" field
        // 4. metadata contains SNS information (TopicArn, MessageId, etc.)

        // Example 2 (Raw SQS) flow:
        // 1. SNSMessageParser.CanParse = false (no SNS properties)
        // 2. EventBridgeMessageParser.CanParse = false (no EventBridge properties)
        // 3. SQSMessageParser.CanParse = true (fallback)
        // 4. messageBody = original message
        // 5. metadata contains just SQS information
        var (messageBody, metadata) = parser.Parse(document.RootElement, sqsMessage);

        // Update the message body if this parser extracted an inner message
        if (!string.IsNullOrEmpty(messageBody))
        {
            // For Example 1:
            // - Updates currentMessageBody to inner message
            // - Creates new JsonElement for next parser to check

            // For Example 2:
            // - This block runs but messageBody is same as original
            currentMessageBody = messageBody;

            document.Dispose(); // Dispose current JsonDocument before reassiginng to parsed message body.
            document = JsonDocument.Parse(messageBody);
        }

        // Combine metadata
        if (metadata.SQSMetadata != null) combinedMetadata.SQSMetadata = metadata.SQSMetadata;
        if (metadata.SNSMetadata != null) combinedMetadata.SNSMetadata = metadata.SNSMetadata;
        if (metadata.EventBridgeMetadata != null) combinedMetadata.EventBridgeMetadata = metadata.EventBridgeMetadata;
    }

    // Example 1 final return:
    // MessageBody = {
    //     "id": "order-123",
    //     "source": "com.myapp.orders",
    //     "type": "OrderCreated",
    //     "time": "2024-03-21T10:00:00Z",
    //     "data": { ... }
    // }
    // Metadata = {
    //     SNSMetadata: { TopicArn: "arn:aws...", MessageId: "abc-123" }
    // }

    // Example 2 final return:
    // MessageBody = {
    //     "id": "order-123",
    //     "source": "com.myapp.orders",
    //     "type": "OrderCreated",
    //     "time": "2024-03-21T10:00:00Z",
    //     "data": { ... }
    // }
    // Metadata = { } // Just basic SQS metadata

    return (currentMessageBody, combinedMetadata);
}
finally
{
    document.Dispose();
}
```

normj (changes requested):

.autover/changes/3347e2c9-b0f5-4145-a8b4-b32b1e6fd511.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Refactor logic for serialization by splitting messaging parsing into multiple classes."
```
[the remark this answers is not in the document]
GarrettBeatty: yeah thats true since there isnt any actual new logic ill remove this change file

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: fixed

normj (approved):

Issue #168, opened 2024-12-18
Not properly formatting the data element of the MessageEnvelope


Scott-transtar:
### Describe the bug

According to the Cloud Event spect, https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md#3-envelope

If the datacontenttype is unspecified (and from what I have seen is not) then SHOULD proceed as if datacontenttype had been specified explicity as application/json.  "The data value MUST be stored directly as a JSON value, rather than as an encoded JSON document represented as a string."

Currently, that is how data is being stored.  Working on either end of the messages, this isn't a problem, however when publishing to an EventBus you cannot reference these values in Event pattern matching.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

A setting is provided to specify the datacontenttype and is respected according to the spec.

If no datacontenttype is specified, as stated in the spec, it should be stored as a JSON value and not an encoded JSON document as a string.

### Current Behavior

data is stored as an encoded JSON document represented as a string.

### Reproduction Steps

Define a POCO for an EventBridge payload, use the AWS.Messaging library to publish the event.  Have an Event Rule defined that will log to a Log Watch group and view the full json of the event, specifically in the detail section.

### AWS.Messaging (or related) package versions

AWS.Messaging-0.9.2

### Targeted .NET Platform

.NET 8

### Operating System and version

Windows 11, any.

bhoradc:
Hi @Scott-transtar, 

Thank you for reporting the issue. Can you please share a minimal reproducible code solution for this issue? 

Regards, 
Chaitanya

Scott-transtar:
Sure, here is a repo that can reproduce the issue.  You will need to supply the ARN of your EventBridge in the Program.cs file.

https://github.com/[Scott-transtar/AWS.Messaging.SerializatinIssue](https://github.com/Scott-transtar/AWS.Messaging.SerializatinIssue)

bhoradc:
Hello @Scott-transtar, 

I appreciate you providing the self-contained code sample, which allows me to replicate the issue quickly.

To reproduce the problem, I followed below steps:

1. Create an Event Bus from AWS console. 
2. For the above Event Bus, create a Rule.
3. Set the Rule type to `Rule with an event pattern`.
4. In the EventBridge pattern, select below pattern to match ANY source
```.json
 {
  "source": [{
    "prefix": ""
  }]
}
```
[...]

PR #193, opened 2025-03-19
Updates the `MessageSerializer` to store the json value as the actual json object in the `data` field.Update data field serialization

GarrettBeatty:
*Issue #, if available:* https://github.com/awslabs/aws-dotnet-messaging/issues/168

*Description of changes:*
1. Updates the `MessageSerializer` to store the json value as the actual json object in the `data` field. 
2. Add `datacontenttype` field to message

**Before**
```
{
    "version": "0",
    "id": "359472c2-c66b-00df-5976-cef04c383a15",
    "detail-type": "Payload",
    "source": "/aws/messaging",
    "account": "147997163238",
    "time": "2025-03-14T18:40:58Z",
    "region": "us-east-1",
    "resources": [],
    "detail": {
        "id": "fdc50c90-f747-4794-ae91-d39894cc6442",
        "source": "/aws/messaging",
        "specversion": "1.0",
        "type": "Payload",
        "time": "2025-03-14T18:40:56.3356898+00:00",
        "data": "{\"Id\":\"8b939477-089c-49ce-ae90-49e9817bd593\",\"Name\":\"Test\",\"SomeValue\":42}"
    }
```

**After**

```
{
    "version": "0",
    "id": "ce7c8a25-da87-879d-100c-874adbaf9c51",
    "detail-type": "Payload",
    "source": "/aws/messaging",
    "account": "147997163238",
    "time": "2025-03-14T19:03:10Z",
    "region": "us-east-1",
    "resources": [],
    "detail": {
        "id": "0ef8cf5e-938a-4dc8-9b62-c6ccb3344bc7",
        "source": "/aws/messaging",
        "specversion": "1.0",
        "type": "Payload",
        "time": "2025-03-14T19:03:08.4022404+00:00",
        "datacontenttype": "application/json",
        "data": {
            "Id": "0172a7de-3757-4740-a771-4aca5f05d67c",
            "Name": "Test",
            "SomeValue": 42
        }
    }
}
```

## Testing
1. I followed the steps in https://github.com/awslabs/aws-dotnet-messaging/issues/168#issuecomment-2563862763 to validate the before and after

### DIFF
--- a/.autover/changes/56a61b2c-4302-4077-9dc6-3fc3bc60a998.json
+++ b/.autover/changes/56a61b2c-4302-4077-9dc6-3fc3bc60a998.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Update MessageSerializer to store data as actual json. Fixes #168"
+      ]
+    }
+  ]
+}
--- a/src/AWS.Messaging/MessageEnvelope.cs
+++ b/src/AWS.Messaging/MessageEnvelope.cs
@@ -45,6 +45,12 @@ public abstract class MessageEnvelope
     public DateTimeOffset TimeStamp { get; set; } = DateTimeOffset.MinValue;
 
     /// <summary>
+    /// The data content type.
+    /// </summary>
+    [JsonPropertyName("datacontenttype")]
+    public string? DataContentType { get; set; }
+
+    /// <summary>
     /// This stores different metadata that is not modeled as a top-level property in MessageEnvelope class.
     /// These entries will also be serialized as top-level properties when sending the message, which
     /// can be used for CloudEvents Extension Attributes.
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -78,8 +78,7 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
             Version = CLOUD_EVENT_SPEC_VERSION,
             MessageTypeIdentifier = publisherMapping.MessageTypeIdentifier,
             TimeStamp = timeStamp,
-            Message = message,
-            DataContentType = _messageSerializer.DataContentType
+            Message = message
         };
     }
 
@@ -103,17 +102,20 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
                 ["source"] = envelope.Source?.ToString(),
                 ["specversion"] = envelope.Version,
                 ["type"] = envelope.MessageTypeIdentifier,
-                ["time"] = envelope.TimeStamp,
-                ["datacontenttype"] = envelope.DataContentType ?? "application/json"
+                ["time"] = envelope.TimeStamp
             };
 
-            if (IsJsonContentType(envelope.DataContentType))
+            var messageSerializerResults = _messageSerializer.Serialize(message);
+
+            blob["datacontenttype"] = messageSerializerResults.ContentType;
+
+            if (IsJsonContentType(messageSerializerResults.ContentType))
             {
-                blob["data"] = JsonNode.Parse(_messageSerializer.Serialize(message));
+                blob["data"] = JsonNode.Parse(messageSerializerResults.Data);
             }
             else
             {
-                blob["data"] = _messageSerializer.Serialize(message);
+                blob["data"] = messageSerializerResults.Data;
 
             }
 
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -78,7 +78,8 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
             Version = CLOUD_EVENT_SPEC_VERSION,
             MessageTypeIdentifier = publisherMapping.MessageTypeIdentifier,
             TimeStamp = timeStamp,
-            Message = message
+            Message = message,
+            DataContentType = _messageSerializer.DataContentType
         };
     }
 
@@ -103,9 +104,18 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
                 ["specversion"] = envelope.Version,
                 ["type"] = envelope.MessageTypeIdentifier,
                 ["time"] = envelope.TimeStamp,
-                ["datacontenttype"] = _messageSerializer.DataContentType,
-                ["data"] = JsonNode.Parse(_messageSerializer.Serialize(message)) // parse the string to get the value as the actual json node.
+                ["datacontenttype"] = envelope.DataContentType ?? "application/json"
             };
+
+            if (IsJsonContentType(envelope.DataContentType))
+            {
+                blob["data"] = JsonNode.Parse(_messageSerializer.Serialize(message));
+            }
+            else
+            {
+                blob["data"] = _messageSerializer.Serialize(message);
+
+            }
 
             // Write any Metadata as top-level keys
             // This may be useful for any extensions defined in
@@ -174,6 +184,34 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
         }
     }
 
+    private bool IsJsonContentType(string? dataContentType)
+    {
+        if (string.IsNullOrWhiteSpace(dataContentType))
+        {
+            // If dataContentType is not specified, it should be treated as "application/json"
+            return true;
+        }
+
+        // Remove any parameters from the content type
+        var mediaType = dataContentType.Split(';')[0].Trim().ToLower();
+
+        // Check if the media type is "application/json"
+        if (mediaType == "application/json")
+        {
+            return true;
+        }
+
+        // Check if the media subtype is "json" or ends with "+json"
+        var parts = mediaType.Split('/');
+        if (parts.Length == 2)
+        {
+            var subtype = parts[1];
+            return subtype == "json" || subtype.EndsWith("+json");
+        }
+
+        return false;
+    }
+
     private  (MessageEnvelope Envelope, SubscriberMapping Mapping) DeserializeEnvelope(string envelopeString)
     {
         using var document = JsonDocument.Parse(envelopeString);
@@ -216,7 +254,10 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
             }
 
             // Deserialize the message content using the custom serializer
-            var dataContent = JsonPropertyHelper.GetRequiredProperty(root, "data", element => element.GetRawText());
+            var dataContent = JsonPropertyHelper.GetRequiredProperty(root, "data", element =>
+                IsJsonContentType(envelope.DataContentType)
+                    ? element.GetRawText()
+                    : element.GetString()!);
             var message = _messageSerializer.Deserialize(dataContent, subscriberMapping.MessageType);
             envelope.SetMessage(message);
 
--- a/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
+++ b/src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
@@ -1,13 +1,12 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-using System.Text;
 using System.Text.Json;
 using System.Text.Json.Nodes;
 using Amazon.SQS.Model;
 using AWS.Messaging.Configuration;
-using AWS.Messaging.Internal;
 using AWS.Messaging.Serialization.Helpers;
+using AWS.Messaging.Internal;
 using AWS.Messaging.Services;
 using Microsoft.Extensions.Logging;
 using AWS.Messaging.Serialization.Parsers;
@@ -104,7 +103,8 @@ internal class EnvelopeSerializer : IEnvelopeSerializer
                 ["specversion"] = envelope.Version,
                 ["type"] = envelope.MessageTypeIdentifier,
                 ["time"] = envelope.TimeStamp,
-                ["data"] = _messageSerializer.Serialize(message)
+                ["datacontenttype"] = _messageSerializer.DataContentType,
+                ["data"] = JsonNode.Parse(_messageSerializer.Serialize(message)) // parse the string to get the value as the actual json node.
             };
 
             // Write any Metadata as top-level keys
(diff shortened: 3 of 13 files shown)


.autover/changes/56a61b2c-4302-4077-9dc6-3fc3bc60a998.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Patch",
```
philasmar, line 5: I think this should be a Minor because there is a significant change in behavior. I would call that out as a breaking change.

src/AWS.Messaging/MessageEnvelope.cs
```diff
@@ -44,6 +44,12 @@ public abstract class MessageEnvelope
     [JsonPropertyName("time")]
     public DateTimeOffset TimeStamp { get; set; } = DateTimeOffset.MinValue;
 
+    /// <summary>
+    /// The data content type.
+    /// </summary>
+    [JsonPropertyName("datacontenttype")]
+    public string? DataContentType { get; set; } = null!;
```
philasmar, line 51: why do you need to explicitly set it to `null!`? The default for `string?` is null.

src/AWS.Messaging/Serialization/IMessageSerializer.cs
```diff
@@ -9,25 +12,24 @@ namespace AWS.Messaging.Serialization;
...
     /// </summary>
-    /// <param name="message">The .NET object that will be serialized.</param>
-    string Serialize(object message);
+    /// <param name="message">The .NET object to be serialized.</param>
+    /// <returns>A serialized representation of the message in a format appropriate for the messaging system.</returns>
+    JsonNode Serialize(object message);
 
     /// <summary>
-    /// Deserializes the raw string message into the .NET type.
+    /// Deserializes a JsonElement message into the specified .NET type.
```
philasmar, line 24: add a doc ref for JsonElement

src/AWS.Messaging/Serialization/IMessageSerializer.cs
```diff
@@ -9,25 +12,24 @@ namespace AWS.Messaging.Serialization;
...
     /// </summary>
-    /// <param name="message">The string message that will be deserialized.</param>
-    /// <param name="deserializedType">The .NET type that represents the deserialized message.</param>
-    object Deserialize(string message, Type deserializedType);
+    /// <param name="message">The JsonElement containing the message to be deserialized.</param>
```
philasmar, line 26: same here

src/AWS.Messaging/Serialization/IMessageSerializer.cs
```diff
@@ -9,25 +12,24 @@ namespace AWS.Messaging.Serialization;
...
-    /// Deserializes the raw string message into the .NET type.
+    /// Gets the content type of the serialized data.
     /// </summary>
-    /// <typeparam name="T">The .NET type that represents the deserialized message.</typeparam>
-    /// <param name="message">The string message that will be deserialized.</param>
-    T Deserialize<T>(string message)
-    {
-        return (T)Deserialize(message, typeof(T));
-    }
+    string DataContentType { get; }
```
philasmar, line 34: Isn't this always "application/json"? How else are we initializing this to a diff val?

src/AWS.Messaging/Serialization/MessageSerializer.cs
```diff
@@ -26,13 +27,22 @@ public MessageSerializer(ILogger<MessageSerializer> logger, IMessageConfiguratio
         _jsonSerializerContext = jsonContextContainer.GetJsonSerializerContext();
     }
 
-    /// <inheritdoc/>
-    /// <exception cref="FailedToDeserializeApplicationMessageException"></exception>
+    /// <summary>
+    /// Deserializes a JsonElement message into the specified type.
```
philasmar, line 31: add doc ref to JsonElement

src/AWS.Messaging/Serialization/MessageSerializer.cs
```diff
@@ -26,13 +27,22 @@ public MessageSerializer(ILogger<MessageSerializer> logger, IMessageConfiguratio
...
+    /// </summary>
+    /// <param name="message">The JsonElement containing the message to deserialize.</param>
```
philasmar, line 33: same here

philasmar (changes requested):

src/AWS.Messaging/Serialization/IMessageSerializer.cs
GarrettBeatty: for our message serializer it will always be application/json. But users can implement their on MessageSerializer if they want (xmlSerializer) or something and have it as a different data content type.

src/AWS.Messaging/Serialization/IMessageSerializer.cs
GarrettBeatty: fixed this and other comments in latest revision

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -205,7 +206,7 @@ private MessageEnvelope DeserializeEnvelope(string envelopeString, Type messageT
             }
 
             // Deserialize the message content using the custom serializer
-            var dataContent = JsonPropertyHelper.GetRequiredProperty(root, "data", element => element.GetString()!);
+            var dataContent = JsonPropertyHelper.GetRequiredProperty(root, "data", element => element);
```
normj, line 209: Can we make this resilient to support the data being either JSON or escaped JSON? I know we are in preview and can make breaking changes but I'm concerned how we are going to roll out this change. If we can make it resilient at least on the reading side that can help.

src/AWS.Messaging/Serialization/IMessageSerializer.cs
```diff
@@ -9,25 +12,24 @@ namespace AWS.Messaging.Serialization;
...
-    /// Serializes the .NET message object into a string.
+    /// Serializes a .NET object into a format suitable for message transport.
+    /// The specific return type depends on the implementation, but must be compatible
+    /// with message transport requirements (e.g., JsonNode for CloudEvents).
     /// </summary>
-    /// <param name="message">The .NET object that will be serialized.</param>
-    string Serialize(object message);
+    /// <param name="message">The .NET object to be serialized.</param>
+    /// <returns>A serialized representation of the message in a format appropriate for the messaging system.</returns>
+    JsonNode Serialize(object message);
```
normj, line 21: This is breaking the abstraction of the service interface to have the Json object from System.Text.Json on the interface. What if a user wants to implement their own message serializer that uses Newtonsoft or not even use JSON.

normj (changes requested):

src/AWS.Messaging/Serialization/IMessageSerializer.cs
GarrettBeatty: I initially used `dynamic` here as the return type instead of `JsonNode` but that was causing trimming issues with NativeAOT. I wasn't sure how to get around that. I believe I also tried using just `object` but that was having issues too. Ill have to double check

GarrettBeatty: the other option is to keep the return type as `object` but then have extra logic in EnvelopeSerializer to handle things but then that defeats the purpose of having everything inside MessageSerializer.

```
           // Handle the data field according to content type
        if (_messageSerializer.DataContentType.Contains("json"))
        {
            // For JSON content, use the serialized JSON value directly
            if (serializedMessage is JsonNode jsonNode)
            {
                blob["data"] = jsonNode;
            }
            else
            {
                // throw exception
            }
        }
        else
        {
            // For non-JSON content, store as string
            blob["data"] = JsonValue.Create(serializedMessage.ToString());
        }
```

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: Hmm. Is there a reason we would want to support both? The correct behavior per the cloudspec is the new logic and feel like since we are still in preview this is the time to make the breaking change. I feel it would complicate the code to support both

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
```diff
@@ -104,7 +103,8 @@ public async ValueTask<string> SerializeAsync<T>(MessageEnvelope<T> envelope)
                 ["specversion"] = envelope.Version,
                 ["type"] = envelope.MessageTypeIdentifier,
                 ["time"] = envelope.TimeStamp,
-                ["data"] = _messageSerializer.Serialize(message)
+                ["datacontenttype"] = _messageSerializer.DataContentType,
+                ["data"] = JsonNode.Parse(_messageSerializer.Serialize(message)) // parse the string to get the value as the actual json node.
```
GarrettBeatty, line 107: this makes the assumption that the user's implementation of message serializer will always return json, which i think is fine for now. Otherwise we have to add more logic in EnvelopeSerializer to handle the case where its not json type for serializing/deserializing. I feel we can add that in the future if we really need it

GarrettBeatty: also @normj responding to your other comment here https://github.com/awslabs/aws-dotnet-messaging/pull/193#discussion_r2006574633  - having a flag to turn on and off for parsing as well would be a similar implementation. so wondering how strongly you feel about that.

philasmar (approved):

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
normj: I don't think we should assume the content return is JSON. Should we have the MessageSerializer's serialize method return back a results type that includes the serialized message and the content type? Then the envelople serializer can look at the content type and decide what to do based on the content type?

GarrettBeatty:
```
return back a results type that includes the serialized message and the content type? Then the envelope serializer can look at the content type and decide what to do based on the content type?
```

im fine with this. i think thats the best way we are going to figure out. i will update the code

GarrettBeatty: I've updated it in the latest revision to have a check in EnvelopeSerializer. And i retested both scenarios manually.

1. Retested with application/json - data gets stored as the actual json object
2. Retested with application/other as the data content type, the data gets stored as the string value

GarrettBeatty: application/json
```
{
    "version": "0",
    "id": "a4579b24-b741-cc2e-df7d-eec8f890da27",
    "detail-type": "Payload",
    "source": "/aws/messaging",
    "account": "147997163238",
    "time": "2025-03-31T17:33:28Z",
    "region": "us-east-1",
    "resources": [],
    "detail": {
        "id": "b08c718c-e63d-4547-9790-030025689103",
        "source": "/aws/messaging",
        "specversion": "1.0",
        "type": "Payload",
        "time": "2025-03-31T17:33:27.1657903+00:00",
        "datacontenttype": "application/json",
        "data": {
            "Id": "036f7898-3e03-403a-b4f6-f18227310330",
            "Name": "Test",
            "SomeValue": 42
        }
    }
}
```
application/plain


```
{
    "version": "0",
    "id": "80f21c60-9797-d928-d0e1-af78e0f7dcba",
    "detail-type": "Payload",
    "source": "/aws/messaging",
    "account": "147997163238",
    "time": "2025-03-31T17:39:49Z",
    "region": "us-east-1",
    "resources": [],
    "detail": {
        "id": "0695f334-bb82-4f03-be12-154cb892b172",
        "source": "/aws/messaging",
        "specversion": "1.0",
        "type": "Payload",
        "time": "2025-03-31T17:39:47.7147488+00:00",
        "datacontenttype": "application/plain",
        "data": "{\"Id\":\"4a6f1d73-289d-4632-88b0-a79318961a25\",\"Name\":\"Test\",\"SomeValue\":42}"
    }
}
````

File: src/AWS.Messaging/Serialization/EnvelopeSerializer.cs [hunk 11]
```diff
@@ -104,9 +104,19 @@ public async ValueTask<string> SerializeAsync<T>(MessageEnvelope<T> envelope)
                 ["specversion"] = envelope.Version,
                 ["type"] = envelope.MessageTypeIdentifier,
                 ["time"] = envelope.TimeStamp,
-                ["data"] = _messageSerializer.Serialize(message)
+                ["datacontenttype"] = envelope.DataContentType ?? "application/json"
             };
 
+            if (IsJsonContentType(envelope.DataContentType))
+            {
+                blob["data"] = JsonNode.Parse(_messageSerializer.Serialize(message));
```
normj, line 112: I was thinking instead of having a global `DataContentType` for the `IMessageSerializer`. The `Serialize` method would return back the content type. That way you could implement different message content type based on the message type. 

I think that would mean the `MessageEnvelope<T>` from `CreateEnvelopeAsync` would leave `DataContentType` null since it is unknown at this point. Then in `SerializeAsync` have something like this:

```csharp
MessageSerializerResults serializationResults = _messageSerializer.Serialize(message);

 var blob = new JsonObject
{
    ["id"] = envelope.Id,
    ["source"] = envelope.Source?.ToString(),
    ["specversion"] = envelope.Version,
    ["type"] = envelope.MessageTypeIdentifier,
    ["time"] = envelope.TimeStamp,
    ["datacontenttype"] = serializationResults.ContentType
};

if (IsJsonContentType(serializationResults.ContentType))
{
    blob["data"] = JsonNode.Parse(serializationResults.Data);
}
else
{
    blob["data"] = serializationResults.Data;
}

class MessageSerializerResults
{
     public string ContentType {get;set;}
     public string Data {get;set;}
}
```

normj (changes requested):

src/AWS.Messaging/Serialization/EnvelopeSerializer.cs
GarrettBeatty: ah okay i see what you mean ! Yes i like that better.  I will update

test/AWS.Messaging.UnitTests/SerializationTests/EnvelopeSerializerTests.cs
```diff
@@ -738,125 +738,23 @@ public async Task ConvertToEnvelope_NullSubscriberMapping_ThrowsException()
...
-    {
-        // ARRANGE
-        var xmlContent = "<root><message>Test XML Content</message></root>";
-        var envelope = new MessageEnvelope<string>
-        {
-            Id = "id-123",
-            Source = new Uri("/backend/service", UriKind.Relative),
-            Version = "1.0",
-            MessageTypeIdentifier = "xml",
-            TimeStamp = _testdate,
```
GarrettBeatty: removed these because testing plain/text was doing same thing

normj (approved):

PR #194, opened 2025-03-19
chore: revert version numbers to match main branch

philasmar:
*Description of changes:*
Since we will only be releasing this library with support for .NET SDK V4, I am reverting the version numbers to match what we currently have on `main`.  Previously, we were planning on releasing a `2.0.0` version for the V4 SDK upgrade which is no longer the case.

### DIFF
--- a/.autover/changes/d7e286d4-e36f-48e9-b6b0-db5ddb200bfa.json
+++ b/.autover/changes/d7e286d4-e36f-48e9-b6b0-db5ddb200bfa.json
@@ -1,0 +1,25 @@
+{
+  "Projects": [
+    {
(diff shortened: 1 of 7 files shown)


GarrettBeatty (approved):

muhammad-othman (approved):

normj (commented):
Should we have some jump in the version number between the versions used in `main` and what is used in `v4` branch? Maybe the v4 branch starts with version number `0.20.0` or something. That way if we have to do some bug fix or something from the v3 main we still do that till we go GA.

normj (approved):

Issue #147, opened 2024-06-21
Feature Request: Start and stop poller?


ashovlin:
### Describe the feature

A potential user reached out internally, looking to pause/resume message handling.  

Our mindset when developing this was that the message pump would run for the whole lifecycle of the application as a `BackgroundService` that would be started by the host.

### Use Case

They use case is for a blue/green deployment where they would just like the active slice to be processing messages.

### Proposed Solution

We do have some spots in our test code where we're starting at stopping it at will, but I think we'd have to flip the visibility of `MessagePumpService` from internal to public before users could do something similar.

[https://github.com/awslabs/aws-dotnet-messaging/blob/ee29c3dadeaf7c49116727db1d2cf[…]2b9ad853a/test/AWS.Messaging.UnitTests/SQSMessagePollerTests.cs](https://github.com/awslabs/aws-dotnet-messaging/blob/ee29c3dadeaf7c49116727db1d2cfb22b9ad853a/test/AWS.Messaging.UnitTests/SQSMessagePollerTests.cs#L269-L279)

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS.Messaging (or related) package versions

N/A

### Targeted .NET Platform

N/A

### Operating System and version

N/A

jkwpappin:
Hi. I have raised a [PR](https://github.com/awslabs/aws-dotnet-messaging/pull/179) for this issue.
Tested via Unit tests. I have an integration test as well but wasn't able to run them locally.

PR #197, opened 2025-03-24
Issue-147: Add feature to pause/resume SQSMessagePoller

philasmar:
*Description of changes:*
Fixing test issues with customer PR https://github.com/awslabs/aws-dotnet-messaging/pull/179 and rebasing on top of `v4sdk-development` branch.
Fixes https://github.com/awslabs/aws-dotnet-messaging/issues/147

### DIFF
--- a/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
+++ b/src/AWS.Messaging/Configuration/IMessageConfiguration.cs
@@ -3,6 +3,7 @@
 
 using AWS.Messaging.Configuration.Internal;
 using AWS.Messaging.Serialization;
+using AWS.Messaging.Services;
 using AWS.Messaging.Services.Backoff;
 using AWS.Messaging.Services.Backoff.Policies;
 using AWS.Messaging.Services.Backoff.Policies.Options;
@@ -93,4 +94,9 @@ public interface IMessageConfiguration
     /// Holds an instance of <see cref="CappedExponentialBackoffOptions"/> to control the behavior of <see cref="CappedExponentialBackoffPolicy"/>.
     /// </summary>
     CappedExponentialBackoffOptions CappedExponentialBackoffOptions { get; }
+
+    /// <summary>
+    /// Holds an instance of <see cref="PollingControlToken"/> to control behaviour of <see cref="IMessagePoller"/>
+    /// </summary>
+    PollingControlToken PollingControlToken { get; }
 }
--- a/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Collections.Concurrent;
 using AWS.Messaging.Configuration.Internal;
 using AWS.Messaging.Publishers;
 using AWS.Messaging.Publishers.EventBridge;
@@ -26,7 +27,7 @@ namespace AWS.Messaging.Configuration;
 /// </summary>
 public class MessageBusBuilder : IMessageBusBuilder
 {
-    private static readonly Dictionary<IServiceCollection, MessageConfiguration> _messageConfigurations = new();
+    private static readonly ConcurrentDictionary<IServiceCollection, MessageConfiguration> _messageConfigurations = new();
     private readonly MessageConfiguration _messageConfiguration;
     private readonly IList<ServiceDescriptor> _additionalServices = new List<ServiceDescriptor>();
     private readonly IServiceCollection _serviceCollection;
--- a/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/MessageBusBuilder.cs
@@ -165,6 +165,13 @@ public class MessageBusBuilder : IMessageBusBuilder
     }
 
     /// <inheritdoc/>
+    public IMessageBusBuilder ConfigurePollingControlToken(PollingControlToken pollingControlToken)
+    {
+        _messageConfiguration.PollingControlToken = pollingControlToken;
+        return this;
+    }
+
+    /// <inheritdoc/>
     [RequiresDynamicCode("This method requires loading types dynamically as defined in the configuration system.")]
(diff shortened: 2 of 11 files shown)


jkwpappin:
Hi @philasmar thanks for fixing the tests! I was just about to look at them, apologies for not looking sooner as it has been a hectic week.

philasmar:
> Hi @philasmar thanks for fixing the tests! I was just about to look at them, apologies for not looking sooner as it has been a hectic week.

Sorry for taking over the PR, thought it'd be a quicker loop if I can run the integ tests through our pipelines and try and figure out what the issue was.

src/AWS.Messaging/Configuration/PollingControlToken.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+namespace AWS.Messaging.Configuration
+{
+    /// <summary>
+    /// Control token to start and stop message polling for a service.
+    /// </summary>
+    public record PollingControlToken
```
normj, line 9: This should be a `class` not a `record`. A record is generally meant for immutable data.

src/AWS.Messaging/Configuration/PollingControlToken.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Stop polling of the SQS Queue.
+        /// </summary>
+        public void StopPolling() => IsPollingEnabled = false;
+
+        /// <summary>
+        /// Configurable amount of time to wait between polling for a change in status
+        /// </summary>
+        public TimeSpan PollingWaitTime { get; set; } = TimeSpan.FromMilliseconds(200);
```
normj, line 29: As this is currently coded different threads are potentially reading and writing to this variable and setting a TimeSpan property is not necessary thread safe. If this needs to be configurable then thread safety has to be taken into account.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -164,6 +165,13 @@ public IMessageBusBuilder AddMessageSourceSuffix(string suffix)
         return this;
     }
 
+    /// <inheritdoc/>
```
normj, line 168: I don't see a change to `IMessageBusBuilder` that adds this new method to the interface? Also make sure the docs you add that is inherited provides more detail about what the PollingControlToken can do besides "control behaviour of <see cref="IMessagePoller"/>". For example can be used for starting and stopping the SQS message poll.

src/AWS.Messaging/Configuration/IMessageConfiguration.cs
```diff
@@ -93,4 +94,9 @@ public interface IMessageConfiguration
     /// Holds an instance of <see cref="CappedExponentialBackoffOptions"/> to control the behavior of <see cref="CappedExponentialBackoffPolicy"/>.
     /// </summary>
     CappedExponentialBackoffOptions CappedExponentialBackoffOptions { get; }
+
+    /// <summary>
+    /// Holds an instance of <see cref="PollingControlToken"/> to control behaviour of <see cref="IMessagePoller"/>
+    /// </summary>
+    PollingControlToken PollingControlToken { get; }
```
normj, line 101: This design is for a single control token across all pollers. If an application is polling multiple queues is it possible they want control this behavior at a per queue level. 

If this is what we want we should call that out that controllers all `IMessagePoller` added via the `AddSQSPoller`. User's are not directly interacting with `IMessagePoller` and might not not what that means so giving the context of `AddSQSPoller` might help with that.

If you do want to control this at a per queue level then maybe this is something passed as part of the `AddSQSPoller` method.

normj (changes requested):

src/AWS.Messaging/Configuration/IMessageConfiguration.cs
jkwpappin: In terms of this feature request in [issue 147](https://github.com/awslabs/aws-dotnet-messaging/issues/147) that we raised, our use-case is for starting/stopping all message consumption at the Service level, not at an individual queue/poller level.

The idea behind the implementation of this is that the `PollingControlToken `gets registered as a Singleton, then I can inject this single `PollingControlToken` into my own background service which will call Start()/Stop() based on it's own logic/conditions.

If you want me to manage or maintain state of multiple `PollingControlToken`s, then you would need to register this by some convention, e.g. KeyedServices, and I'd have to maintain some kind of additional state of which `PollingControlToken `matches which `IMessagePoller`. I feel this makes the client usage of these more complicated for the Service level start/stop use-case. Alternatively you could provide a mechanism where I can ask the MessagingFramework to give me the `PollingControlToken `for a specific `IMessagePoller`, but again I would need to maintain state/view of which `IMessagePoller `I have and want to stop. We could also add a mechanism to the framework where it can start/stop all registered PollingControlToken's, might be the best option if you want to also allow control on individual `IMessagePoller `instances.

normj: Looking at this feature with ignorance I wasn't sure what the use cases were and if it made sense to control start/stop at a per queue or global level. It sounds like you are saying the use case for a per scenario is not what you are looking for and its complexity. I'm fine not doing it and keeping it global, just wanted to ask the question.

src/AWS.Messaging/Configuration/PollingControlToken.cs
normj: I would be fine changing this to an `init` to avoid the thread complexity.

src/AWS.Messaging/Configuration/PollingControlToken.cs
philasmar: Fixed

src/AWS.Messaging/Configuration/PollingControlToken.cs
philasmar: Updated

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
philasmar: Added

GarrettBeatty (approved):

normj (approved):

PR #209, opened 2025-04-02
Fix null reference exception

GarrettBeatty:
*Issue #, if available:*

*Description of changes:*
When we poll SQS and no messages are received, `messages` is `null` but we are still trying to access the count anyway. This adds a check to ensure its not null

image


normj (approved):
I just ran into this as well when testing your sample app PR.

GarrettBeatty:
> I just ran into this as well when testing your sample app PR.

yep thats how i found it too!

philasmar (approved):

PR #249, opened 2025-05-28
Fix OTEL for processing message

GarrettBeatty:
*Issue #, if available:*
https://github.com/aws/aws-dotnet-messaging/issues/159

*Description of changes:*

This change changes the behavior of how the consumer otel processing works. Previously we were setting the consumer as a child span of the producer. This was causing issues when viewing in X-ray/looking at traces, where the publisher and consumer were shown as one node (see picture below)

X-ray
333184543-27e810af-e3c3-4d89-9ff8-f64c3afd1231

Jaeger
Pasted image 20250527104230

## Testing in Xray
1. Publisher -> SQS -> ECS (consumer)

Pasted image 20250527143428

2. Publisher -> SQS -> Lambda (consumer)

Pasted image 20250528114012

You can see in each of these pictures that the producer and consumer are separate nodes.

### Testing in Jaeger

Pasted image 20250528115254
Pasted image 20250528115351

You can see there is a linked span. And when its clicked it takes you to the publisher
Pasted image 20250528115430

## Notes

image
In this current implementation there is  UnknownRemoteService and Operation value and this is due to not setting some of the xray/aws specific values https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Enable-Troubleshoot.html#Application-Signals-troubleshoot-unknown-values. I dont know if it makes sense to set these those (or what to put if we want to) because at the point of creating the span, we aren't actually sending to SQS. There is no "remote operation" at that point. If the user wanted to see the remote operation, it  automatically be captured (in a different span) if the user uses AddAWSAutoInstrumentation

### DIFF
--- a/.autover/changes/bebdd3cf-553b-4a09-a5d6-59af70e2c5ff.json
+++ b/.autover/changes/bebdd3cf-553b-4a09-a5d6-59af70e2c5ff.json
@@ -2,7 +2,7 @@
   "Projects": [
     {
       "Name": "AWS.Messaging.Telemetry.OpenTelemetry",
-      "Type": "Patch",
+      "Type": "Minor",
       "ChangelogMessages": [
         "Updated consumer processing logic to use Links for connecting to publisher instead of parent-child relationship"
       ]
--- a/.autover/changes/bebdd3cf-553b-4a09-a5d6-59af70e2c5ff.json
+++ b/.autover/changes/bebdd3cf-553b-4a09-a5d6-59af70e2c5ff.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging.Telemetry.OpenTelemetry",
+      "Type": "Patch",
+      "ChangelogMessages": [
+        "Updated consumer processing logic to use Links for connecting to publisher instead of parent-child relationship"
+      ]
+    }
+  ]
+}
--- a/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
+++ b/src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
@@ -34,7 +34,7 @@ public class OpenTelemetryProvider : ITelemetryProvider
         }
 
         activity = _activitySource.StartActivity(traceName, ActivityKind.Producer, parentContext: default, links: links);
-        
+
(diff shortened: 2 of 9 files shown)


src/AWS.Messaging.Telemetry.OpenTelemetry/OpenTelemetryProvider.cs
```diff
@@ -44,23 +44,43 @@ public ITelemetryTrace Trace(string traceName, MessageEnvelope envelope)
...
+
         if (activity != null)
         {
+            activity.SetStatus(ActivityStatusCode.Ok);
             return new OpenTelemetryTrace(activity);
         }
 
-        // If we initially failed to create an activity, attempt to force creation with 
-        // a link to the current activity, see https://opentelemetry.io/docs/instrumentation/net/manual/#creating-new-root-activities
+        // If we initially failed to create an activity, attempt to force creation
```
GarrettBeatty, line 68: i dont know if this is actaully needed or not but they had it originally so i kept it

normj:
@gcbeatty Reading your comment about the UnknownServices would be filled in if the `AddAWSAutoInstrumentation` for the SDK package was added. I'm wondering if `AWS.Messaging.Telemetry.OpenTelemetry` should take a dependency on `OpenTelemetry.Instrumentation.AWS` and we call `AddAWSAutoInstrumentation` as part of `AddAWSMessagingInstrumentation`.

I think then we could explain how `AWS.Messaging.Telemetry.OpenTelemetry` builds on top of `OpenTelemetry.Instrumentation.AWS` adding MPF context information. I assume the `AddAWSAutoInstrumentation` method is safe to call twice in case the user also calls it. If not we should make it so it is safe to call more then once.

.autover/changes/bebdd3cf-553b-4a09-a5d6-59af70e2c5ff.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging.Telemetry.OpenTelemetry",
+      "Type": "Patch",
```
normj, line 5: nit: I would suggest a minor update given we are making a pretty big change to how telemetry is connected.

normj (approved):
Minor suggestion for the change file but PR looks good.

Issue #273, opened 2025-07-15
Reduce Allocations in Envelope Serialization


JanHyka:
### Describe the feature

The current implementation of `EnvelopeSerializer` utilizes `JsonNode` for message serialization. While functional, this approach introduces unnecessary allocations and increased memory usage, particularly in high-throughput scenarios. There is an opportunity to significantly reduce allocations and improve performance by leveraging `Utf8JsonWriter` over rented buffers, thereby enabling buffer reuse and minimizing memory overhead per message.

### Use Case

Serialization of any enveloped payload passing through the library.

#### Replace `JsonNode` Usage:
Transition the serialization process from `JsonNode` to `Utf8JsonWriter`, utilizing rented memory buffers. This change will reduce object allocations and improve memory management, especially under heavy load.

#### Interface Compatibility:
To maintain backward compatibility, the `EnvelopeSerializer.SerializeAsync` method should continue to return a string. In the future, further optimizations could be realized by returning a buffer directly, but such a change would introduce a breaking change to the API.

#### Potential for Future Improvements:
Passing a buffer to `EnvelopeSerializer.SerializeAsync` would allow dropping string allocations. With possible long term solution entirely working on rented buffers only across all layers.

#### Introduce Non-Allocating Serialization:
The current `IMessageSerializer` interface’s `MessageSerializerResult Serialize` method is incompatible with non-allocating serialization scenarios. To address this, a new method is proposed:

`
ValueTask<string> SerializeToBufferAsync(Utf8JsonWriter writer, object? value, ...);
`
where return value is content type.

* Note: Evaluate whether it is feasible to utilize IBufferedWriter<byte> (to avoid tight coupling to Utf8JsonWriter) without disrupting the typical Utf8JsonWriter usage patterns.

#### Incremental Interface Introduction:
To ensure a smooth transition, introduce a new interface, `IJsonWriterMessageSerializer`, which derives from `IMessageSerializer` and adds the proposed non-allocating serialization method. This approach allows incremental adoption without breaking existing consumers.

`
public interface IJsonWriterMessageSerializer : IMessageSerializer
{
    ValueTask<string> SerializeToBufferAsync(Utf8JsonWriter writer, object? value, ...);
}
`

#### Runtime Detection and Adaptation:
[...]

PR #280, opened 2025-08-16
Reduce Allocations in Envelope Serialization

JanHyka:
*Issue #, if available:*
#273 Reduce Allocations in Envelope Serialization

*Description of changes:*

- new version of `IEnvelopeSerializer` was introduced: `EnvelopeSerializerUtf8JsonWriter`
- serializer method was rewritten using optimized approach based on Utf8JsonWriter
- new version of `IMesssageSerializer` was introduced: `MessageSerializerUtf8JsonWriter`
- necessary interface changes were added via `IMessageSerializerUtf8JsonWriter` interface
- `MessageBusBuilder` was extended by `EnableExperimentalFeatures()` method - when used, new Envelope/Message serializers are used
- `SerializationOptions` was extended by `bool CleanRentedBuffers` (by default true). At cost of security risk keeping uncleared buffers in memory makes the result a bit faster (**suggestion here is to drop the option and clean always**)
- Unit tests were added to cover the changes/newly added code
- `AWS.Messaging.Benchmarks.Serialization` project was added for convenience (**can be kept or dropped based on preferences**)

*Benchmark results:*
```

BenchmarkDotNet v0.13.10, Windows 11 (10.0.26100.4652)
Unknown processor
.NET SDK 10.0.100-preview.2.25164.34
  [Host]     : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX2


```
| Method                                    | ItemCount | Mean         | Error     | StdDev    | Median       | Ratio | RatioSD | Gen0    | Gen1    | Gen2    | Allocated | Alloc Ratio |
|------------------------------------------ |---------- |-------------:|----------:|----------:|-------------:|------:|--------:|--------:|--------:|--------:|----------:|------------:|
| **StandardSerializer**                        | **1**         |   **1,047.7 ns** |   **9.08 ns** |   **8.49 ns** |   **1,047.1 ns** |  **1.00** |    **0.00** |  **0.1240** |       **-** |       **-** |    **2368 B** |        **1.00** |
| StandardSerializerWithJsonContext         | 1         |     941.7 ns |   2.85 ns |   2.67 ns |     942.0 ns |  0.90 |    0.01 |  0.1087 |       - |       - |    2056 B |        0.87 |
| JsonWriterSerializer                      | 1         |     458.4 ns |   2.27 ns |   2.12 ns |     458.6 ns |  0.44 |    0.00 |  0.0548 |       - |       - |    1040 B |        0.44 |
| JsonWriterSerializerWithJsonContext       | 1         |     357.0 ns |   1.71 ns |   1.60 ns |     357.4 ns |  0.34 |    0.00 |  0.0386 |       - |       - |     728 B |        0.31 |
| JsonWriterSerializerWithJsonContextUnsafe | 1         |     276.4 ns |   0.64 ns |   0.57 ns |     276.3 ns |  0.26 |    0.00 |  0.0386 |       - |       - |     728 B |        0.31 |
|                                           |           |              |           |           |              |       |         |         |         |         |           |             |
| **StandardSerializer**                        | **10**        |   **3,162.9 ns** |  **19.17 ns** |  **17.93 ns** |   **3,168.6 ns** |  **1.00** |    **0.00** |  **0.2861** |       **-** |       **-** |    **5432 B** |        **1.00** |
| StandardSerializerWithJsonContext         | 10        |   2,617.0 ns |   6.00 ns |   5.61 ns |   2,616.1 ns |  0.83 |    0.01 |  0.2708 |       - |       - |    5120 B |        0.94 |
| JsonWriterSerializer                      | 10        |   1,112.1 ns |   5.10 ns |   4.26 ns |   1,112.6 ns |  0.35 |    0.00 |  0.1011 |       - |       - |    1920 B |        0.35 |
| JsonWriterSerializerWithJsonContext       | 10        |     696.2 ns |   2.06 ns |   1.92 ns |     695.8 ns |  0.22 |    0.00 |  0.0849 |       - |       - |    1608 B |        0.30 |
| JsonWriterSerializerWithJsonContextUnsafe | 10        |     598.4 ns |   4.37 ns |   4.09 ns |     599.8 ns |  0.19 |    0.00 |  0.0849 |       - |       - |    1608 B |        0.30 |
|                                           |           |              |           |           |              |       |         |         |         |         |           |             |
| **StandardSerializer**                        | **100**       |  **27,274.1 ns** | **491.14 ns** | **820.58 ns** |  **27,431.5 ns** |  **1.00** |    **0.00** |  **1.9531** |  **0.1526** |       **-** |   **37032 B** |        **1.00** |
| StandardSerializerWithJsonContext         | 100       |  19,778.3 ns | 389.58 ns | 702.49 ns |  19,639.0 ns |  0.73 |    0.04 |  1.9226 |  0.1526 |       - |   36720 B |        0.99 |
| JsonWriterSerializer                      | 100       |   7,880.9 ns | 157.43 ns | 303.32 ns |   7,744.5 ns |  0.29 |    0.01 |  0.5798 |       - |       - |   11104 B |        0.30 |
| JsonWriterSerializerWithJsonContext       | 100       |   4,023.4 ns |  80.33 ns | 222.61 ns |   3,955.4 ns |  0.15 |    0.01 |  0.5722 |       - |       - |   10792 B |        0.29 |
| JsonWriterSerializerWithJsonContextUnsafe | 100       |   3,617.6 ns |  35.00 ns |  32.74 ns |   3,600.0 ns |  0.13 |    0.01 |  0.5722 |       - |       - |   10792 B |        0.29 |
|                                           |           |              |           |           |              |       |         |         |         |         |           |             |
| **StandardSerializer**                        | **1000**      | **274,182.0 ns** | **752.42 ns** | **667.00 ns** | **274,142.0 ns** |  **1.00** |    **0.00** | **96.6797** | **96.6797** | **96.6797** |  **361993 B** |        **1.00** |
| StandardSerializerWithJsonContext         | 1000      | 238,880.0 ns | 441.12 ns | 391.04 ns | 238,938.5 ns |  0.87 |    0.00 | 96.6797 | 96.6797 | 96.6797 |  361681 B |        1.00 |
| JsonWriterSerializer                      | 1000      | 101,512.4 ns | 260.47 ns | 243.65 ns | 101,461.8 ns |  0.37 |    0.00 | 33.3252 | 33.3252 | 33.3252 |  106538 B |        0.29 |
| JsonWriterSerializerWithJsonContext       | 1000      |  69,257.1 ns | 267.03 ns | 222.98 ns |  69,241.4 ns |  0.25 |    0.00 | 33.3252 | 33.3252 | 33.3252 |  106226 B |        0.29 |
| JsonWriterSerializerWithJsonContextUnsafe | 1000      |  69,293.8 ns | 320.33 ns | 299.64 ns |  69,175.4 ns |  0.25 |    0.00 | 33.3252 | 33.3252 | 33.3252 |  106226 B |        0.29 |


src/AWS.Messaging/Serialization/EnvelopeSerializerUtf8JsonWriter.cs
```diff
@@ -0,0 +1,473 @@
...
+            await InvokePreSerializationCallback(envelope);
+            T message = envelope.Message ?? throw new ArgumentNullException("The underlying application message cannot be null");
+
+            using var buffer = new RentArrayBufferWriter(cleanRentedBuffers: _messageConfiguration.SerializationOptions.CleanRentedBuffers);
+            using var writer = new Utf8JsonWriter(buffer, s_serializerWriterOptions);
+
+            writer.WriteStartObject();
+
+            writer.WriteString(s_idProp, envelope.Id);
+            writer.WriteString(s_sourceProp, envelope.Source?.OriginalString);
```
GarrettBeatty, line 118: i think in the regular enevelopserializer we do ToString() . is there any difference between that vs original string? was seeing this when searching online

```
ToString():
This method returns a canonicalized and unescaped representation of the URI. This means it might perform actions like converting the scheme and host to lowercase, removing default port numbers, and unescaping percent-encoded characters (like %20 becoming a space). The goal is to provide a "human-readable" or standardized form of the URI.
OriginalString:
This property returns the exact string that was used to initialize the Uri object, without any modifications, canonicalization, or unescaping. If the original string contained leading/trailing spaces, non-standard casing, or percent-encoded characters, OriginalString will preserve them precisely as they were provided.
```

can we verify?

src/AWS.Messaging/Serialization/EnvelopeSerializerUtf8JsonWriter.cs
```diff
@@ -0,0 +1,473 @@
...
+                _messageSerializerUtf8Json.SerializeToBuffer(writer, message);
+            }
+            else
+            {
+                var response = _messageSerializer.Serialize(message);
+                writer.WriteString(s_dataContentTypeProp, response.ContentType);
+                writer.WritePropertyName(s_dataProp);
+                if (IsJsonContentType(response.ContentType))
+                {
+                    writer.WriteRawValue(response.Data, skipInputValidation: true);
```
GarrettBeatty, line 136: why is it safe to skip validation?

test/AWS.Messaging.UnitTests/SerializationTests/MessageSerializerTestsUtf8JsonWriterTests.cs
```diff
@@ -0,0 +1,432 @@
...
+using AWS.Messaging.UnitTests.Models;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Logging.Testing;
+using Moq;
+using Xunit;
+
+namespace AWS.Messaging.UnitTests.SerializationTests;
+
+public class MessageSerializerUtf8JsonWriterTests
```
GarrettBeatty, line 20: question: for these test cases do we expect them to be equivalent to the message serialize test cases? (i.e the same input will give same output?) i was thinking the answer is "yes" because the the new message serializer would only have changes related to how it creates the results (with all of its optimizations and everything) but the final json should be the same. 

if thats the case, wondering why we didnt do similar to envelope serializer, where we run the same test cases against both scenarios and they produce identical results?

src/AWS.Messaging/Serialization/EnvelopeSerializerUtf8JsonWriter.cs
JanHyka: my reasoning is following:

- we have full control over the envelope json structure and property names (they are pre-encoded anyways)
- Utf8JsonWriter handles escaping for all string writing calls internally
- metadata get emitted via JsonElement.WriteTo which guarantees valid JSON
- for data either we have Utf8Serializer that writes directly and safely or fallback to non-utf ones. Out of those, the ones returning non-JSON go via string writing route and for other we rely on message serializer returning valid json for its advertised content type

src/AWS.Messaging/Serialization/EnvelopeSerializerUtf8JsonWriter.cs
JanHyka: Also, another point of view can be when checking to Uri implementation:

```
public bool IsAbsoluteUri { get { return _syntax != null }}
```
so, when Uri is not absolute, which it is not in our case,

```
public override string ToString()
{
    if (_syntax == null)
    {
        return _string;
    }
    ...
```
and
```
//
//  Gets the exact string passed by a user.
//  The original string will switched from _string to _originalUnicodeString if
//  iri is turned on and we have non-ascii chars
//
public string OriginalString => _originalUnicodeString ?? _string;
```

will be returning exactly same thing, we can easily keep ToString() if you prefer as in our use case, both will be and behave exactly same.

GarrettBeatty (approved):

GarrettBeatty:
https://github.com/aws/aws-dotnet-messaging/actions/runs/21256221138 ci run

src/AWS.Messaging/Configuration/SerializerOptions.cs
```diff
@@ -22,6 +22,14 @@ public class SerializationOptions
         Converters = { new JsonStringEnumConverter() }
     };
 
+    /// <summary>
+    /// When set to true, it will clean the rented buffers after each use.
+    /// </summary>
+    /// <remarks>
+    /// Setting this to false can improve performance in high-throughput scenarios at cost of potential security issues
+    /// </remarks>
+    public bool CleanRentedBuffers { get; set; } = true;
```
normj: A problem with this flag is it only applies if you are using the new serializer. Users might not realize that and are using the original serializer and this flag being true implies security a promise might not be keeping. Not sure what the JsonSerializer.Serialize method does for this in the original serializer. It goes to my comment elsewhere that if this is right serializer to use we should aim to get rid of the EnableExperimental method and switch over to it.

src/AWS.Messaging/Serialization/Helpers/RentArrayBufferWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Buffers;
+using System.Diagnostics;
+
+/// <summary>
+/// https://gist.github.com/ahsonkhan/c76a1cc4dc7107537c3fdc0079a68b35
+/// Standard ArrayBufferWriter is not using pooled memory
+/// </summary>
+internal class RentArrayBufferWriter : IBufferWriter<byte>, IDisposable
+{
+    private const int MINIMUM_BUFFER_SIZE = 256;
```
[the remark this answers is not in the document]
normj: I admit I don't have good data to base a decision on but I do believe messages are most commonly at least 2 to 10 KB looking anecdotally internally. The SDK generally uses an 8KB buffer when reading from streams. What about starting with 2 KB and exposing this as a configuration knob for users that anticipate having much larger messages.

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -321,6 +323,12 @@ public IMessageBusBuilder ConfigureBackoffPolicy(Action<BackoffPolicyBuilder> co
         return this;
     }
 
+    public IMessageBusBuilder EnableExperimentalFeatures()
```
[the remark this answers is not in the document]
JanHyka: ack, I'll update the PR with your inputs and prepare cleaned out ready-for-merge version.

src/AWS.Messaging/Configuration/SerializerOptions.cs
JanHyka: Since we are also adding configurable initial buffer size I have incorporated both changes to RentedBufferOptions.

GarrettBeatty:
@normj can you review this please

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
```diff
@@ -376,8 +376,11 @@ internal void Build()
 
         _serviceCollection.TryAddSingleton(_messageConfiguration.PollingControlToken);
         _serviceCollection.TryAddSingleton<IMessageConfiguration>(_messageConfiguration);
+        
+        _serviceCollection.AddSingleton<IEnvelopeSerializer, EnvelopeSerializer>();
         _serviceCollection.TryAddSingleton<IMessageSerializer, MessageSerializer>();
-        _serviceCollection.TryAddSingleton<IEnvelopeSerializer, EnvelopeSerializer>();
```
AlexDaines, line 380:
```csharp
_serviceCollection.AddSingleton<IEnvelopeSerializer, EnvelopeSerializer>();
_serviceCollection.TryAddSingleton<IMessageSerializer, MessageSerializer>();
```

On `main` both lines are `TryAddSingleton`. The `AddSingleton` looks like leftover from when the experimental path registered a second `IEnvelopeSerializer` implementation; after collapsing to a single class it should go back to `TryAddSingleton` to match the surrounding pattern and avoid double-registration if `Build()` runs more than once.

This is the only behavioral regression I'd block on. It is also the same line Copilot/Garrett flagged earlier in the thread (then about a duplicated registration); the duplicate is gone but the `Try` is still missing.

src/AWS.Messaging/Serialization/IMessageSerializerUtf8JsonWriter.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Text.Json;
+
+namespace AWS.Messaging.Serialization;
+
+/// <summary>
+/// Supports serialization and deserialization of domain-specific application messages.
+/// This interface extends <see cref="IMessageSerializer"/> to provide a methods for allocation-free serialization/deserialization.
+/// </summary>
+public interface IMessageSerializerUtf8JsonWriter
```
AlexDaines: normj's request: "make `IMessageSerializerUtf8JsonWriter` an `internal` interface and treat this as an internal optimization. That allows us to punt for later decisions on further extensibility on this if/when we get asks for that." JanHyka acked ("ack, I'll update the PR with your inputs and prepare cleaned out ready-for-merge version"), but the interface is still declared `public`. Either flip to `internal`, or surface explicitly that the decision changed.

If kept public: doc on `SerializeToBuffer<T>` says "allocation-free serialization" — it is *lower* allocation, not allocation-free (the `JsonSerializer.Serialize(writer, ...)` call still allocates internally for non-trivial graphs).

AlexDaines (changes requested):

src/AWS.Messaging/Configuration/MessageBusBuilder.cs
JanHyka: Good catch, yes, this wasn't supposed to be plain Add.

src/AWS.Messaging/Serialization/IMessageSerializerUtf8JsonWriter.cs
JanHyka: I have changed the interface to internal plus changed the description to just point the usage of rented buffers. The logic behind it is more complex and while it is commonly called allocation-less, there indeed are cases when rented buffers do allocate more memory.

normj (approved):

GarrettBeatty:
https://github.com/aws/aws-dotnet-messaging/actions/runs/25750920580

GarrettBeatty:
merged the latest dev branch code into this branch which should fix the build

GarrettBeatty:
https://github.com/aws/aws-dotnet-messaging/actions/runs/25752674334

GarrettBeatty:
Thanks for the contribution @JanHyka

PR #311, opened 2026-01-22
Update README with AWS Event info

philasmar:
*Issue #, if available:*
DOTNET-7590

*Description of changes:*
Add a section to describe how the framework can be used with AWS events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,78 @@ await Host.CreateDefaultBuilder(args)
     .RunAsync();
 ```
 
+### Consuming AWS service events (raw JSON payloads)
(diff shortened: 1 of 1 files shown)


normj (approved):
Approving assuming you have verified the Lambda event objects are compatible with the JSON that comes into SQS directly from the service.

normj (changes requested):
Change approval because the PR is suggesting using the Lambda SQS package but that doesn't make sense.

GarrettBeatty (approved):

normj (approved):

PR #320, opened 2026-03-30
Add ISerializationCallback<T> for type-specific serialization callbacks

GarrettBeatty:
## Description

Adds a new `ISerializationCallback<T>` generic interface for type-specific serialization callbacks. Callbacks are only invoked when the message being serialized matches `T`, providing **direct typed access to the message payload with zero casting required**. This addresses the challenge described in [Discussion #317](https://github.com/aws/aws-dotnet-messaging/discussions/317).

Also fixes `ArgumentNullException` constructor misuse in `EnvelopeSerializer` where the error message was passed as `paramName`.

### Problem

Users want to set CloudEvents extension attributes (like `subject`) by extracting values from the message payload. The existing `ISerializationCallback.PreSerializationAsync(MessageEnvelope)` only provides the abstract `MessageEnvelope`, forcing users to cast:

```csharp
// Before — user must cast to access the message
public class MyCallback : ISerializationCallback
{
    public ValueTask PreSerializationAsync(MessageEnvelope messageEnvelope)
    {
        if (messageEnvelope is MessageEnvelope<UserCreatedEvent> typed) // casting required!
        {
            typed.Metadata["subject"] = JsonSerializer.SerializeToElement(typed.Message.UserId);
        }
        return ValueTask.CompletedTask;
    }
}
```

### Solution

New `ISerializationCallback<T>` interface following the same pattern as `IMessageHandler<T>`:

```csharp
public interface ISerializationCallback<T>
{
    ValueTask PreSerializationAsync(MessageEnvelope<T> messageEnvelope);
}
```

Callbacks are registered via DI and automatically type-routed — they're only invoked when `T` matches:

```csharp
// After — zero casting, direct typed access, DI-resolved
public class UserCreatedCallback : ISerializationCallback<UserCreatedEvent>
{
    public ValueTask PreSerializationAsync(MessageEnvelope<UserCreatedEvent> messageEnvelope)
    {
        // Direct typed access — no casting!
        messageEnvelope.Metadata["subject"] = 
            JsonSerializer.SerializeToElement(messageEnvelope.Message.UserId);
        return ValueTask.CompletedTask;
    }
}

// Registration
services.AddAWSMessageBus(builder =>
{
    builder.AddSQSPublisher<UserCreatedEvent>("queue-url", "user.created");
    builder.AddSerializationCallback<UserCreatedCallback, UserCreatedEvent>();
});
```

### How It Works

1. **Registration**: `AddSerializationCallback<TCallback, TMessage>()` registers the callback in the DI container as `ISerializationCallback<TMessage>`
2. **Invocation**: In `EnvelopeSerializer.SerializeAsync<T>()`, after invoking non-generic callbacks, it calls `_serviceProvider.GetServices<ISerializationCallback<T>>()` — the DI container returns **only** callbacks matching `T`
3. **Type safety**: No reflection needed. Since `SerializeAsync<T>` knows `T` at compile time, the DI container handles all type routing automatically

### Produces the Output the User Wanted

```json
{
  "type": "user.created",
  "source": "my-service",
  "subject": "123",
  "data": { "userId": "123" }
}
```

### Changes

| File | Change |
|------|--------|
| `ISerializationCallback{T}.cs` | **New** — generic typed callback interface |
| `IMessageBusBuilder.cs` | Added `AddSerializationCallback<TCallback, TMessage>()` |
| `MessageBusBuilder.cs` | DI registration for typed callbacks |
| `EnvelopeSerializer.cs` | Added `IServiceProvider` dep, typed callback invocation, fixed `ArgumentNullException` |
| `ISerializationCallback.cs` | Reverted previous generic method (replaced by this approach) |
| `EnvelopeSerializerTests.cs` | New typed callback tests, updated direct constructor calls |

### Breaking Changes

None. The existing `ISerializationCallback` (non-generic) continues to work unchanged for cross-cutting concerns.

### Testing

- All 20 existing EnvelopeSerializer tests pass
- 2 new tests:
  - `TypedSerializationCallback_ExtractsSubjectWithZeroCasting` — verifies typed callback extracts subject with no casting
  - `TypedSerializationCallback_NotInvokedForNonMatchingType` — verifies callback is NOT invoked for non-matching message types

*Issue #, if available:* Resolves discussion #317

### DIFF
--- a/.autover/changes/8182ce75-83d8-465c-b65c-d29f7b731729.json
+++ b/.autover/changes/8182ce75-83d8-465c-b65c-d29f7b731729.json
@@ -4,7 +4,7 @@
       "Name": "AWS.Messaging",
       "Type": "Minor",
       "ChangelogMessages": [
-        "Added generic PreSerializationAsync\u003CT\u003E(MessageEnvelope\u003CT\u003E) overload to ISerializationCallback, allowing typed access to the message payload during pre-serialization callbacks. This enables extracting values from the message to set as CloudEvents extension attributes (e.g., subject)."
+        "Added ISerializationCallback\u003CT\u003E generic interface for type-specific serialization callbacks. Callbacks registered via AddSerializationCallback\u003CTCallback, TMessage\u003E() are only invoked when the message type matches, providing direct typed access to the message payload with zero casting. Also fixed ArgumentNullException constructor usage in EnvelopeSerializer."
       ]
     }
   ]
--- a/.autover/changes/8182ce75-83d8-465c-b65c-d29f7b731729.json
+++ b/.autover/changes/8182ce75-83d8-465c-b65c-d29f7b731729.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Added generic PreSerializationAsync\u003CT\u003E(MessageEnvelope\u003CT\u003E) overload to ISerializationCallback, allowing typed access to the message payload during pre-serialization callbacks. This enables extracting values from the message to set as CloudEvents extension attributes (e.g., subject)."
+      ]
+    }
+  ]
+}
--- a/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
+++ b/src/AWS.Messaging/Configuration/IMessageBusBuilder.cs
@@ -92,6 +92,16 @@ public interface IMessageBusBuilder
     IMessageBusBuilder AddSerializationCallback(ISerializationCallback serializationCallback);
 
     /// <summary>
+    /// Adds a type-specific serialization callback that is only invoked when the message being
+    /// serialized matches <typeparamref name="TMessage"/>. This provides direct typed access to
+    /// the message payload without requiring any casting.
+    /// </summary>
+    /// <typeparam name="TCallback">The callback implementation type.</typeparam>
+    /// <typeparam name="TMessage">The message type this callback handles.</typeparam>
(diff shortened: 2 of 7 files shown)


normj (changes requested):
If a user still has to do casting like `if (messageEnvelope.Message is IHasSubject subjectProvider)` in the new generic solution our we really making things much better then the current implementation where the user could cast to ` if (messageEnvelope is MessageEnvelope<AddressInfo> addressInfoEnvelope)`.


If we want to make this easier I wonder if the user should be able to register a concrete generic like `ISerializationCallback<AddressInfo>` and the callback is only called if the type being serialized matches the generic type. Then the user wouldn't have to know to do any casting.

normj (approved):

philasmar (approved):

abc-rpg:
@GarrettBeatty Thanks for the looking into this. Unfortunately, I just realized that the revised implementation does not work for my requirement, while the first iteration did.

The issue is with a message interface or base class.

```csharp
public interface IMessageWithSubject
{
    public string GetSubject();
}

public class SetMessageSubjectCallback : ISerializationCallback<IMessageWithSubject>
{
    public ValueTask PreSerializationAsync(MessageEnvelope<IMessageWithSubject> envelope)
    {
        envelope.Metadata["subject"] = JsonSerializer.SerializeToElement(envelope.Message.GetSubject());
    }
}

// Program.cs
services.AddAWSMessageBus(builder => builder.AddSerializationCallback<SetSubjectMessageCallback, IMessageWithSubject>();
```

When I then publish some mesasge `MyMessage` that extends `IMessageWithSubject`, the preserialization callback will not fire because it will look for `ISerializationCallback<MyMessage>`.

I would need to implement an `ISerializationCallback` for each unique message type - to execute the exact same logic.

The first implementation that required type casting on my side, allowed me to check if any message implemented the base message type - which was enough to extract the subject.

Should I open a new issue/discussion for this?

Issue #340, opened 2026-05-13
Publish to FIFO queue with IMessagePublisher


EtherZa:
### Describe the feature

Per message options configuration for SQS/SNS messages when using `IMessagPublisher`.

### Use Case

To support FIFO queues where there is sufficient detail in the message to determine the group/deduplication id's. Having the configuration at bootstrap abstracts the FIFO logic/requirement away from the code that publishes the message.

### Proposed Solution

Add optional configuration actions to `IMessageBusBuilder.AddSQSPublisher` and `IMessageBusBuilder.AddSNSPublisher` for `MessageRoutingPublisher` to pre-populate `SQSOptions`/`SNSOptions` before handing off to the specialised publishers.

```csharp
builder.Services.AddAWSMessageBus(builder =>
{
    builder.AddSQSPublisher<TransactionInfo>(
        "https://sqs.us-west-2.amazonaws.com/012345678910/MyFifoQueue.fifo",
        configureOptions: (message, options) =>
        {
            options.MessageGroupId = message.TransactionId;
            options.MessageDeduplicationId = message.TransactionId;
        });

    builder.AddSNSPublisher<BidInfo>(
        "arn:aws:sns:us-west-2:012345678910:MyFifoTopic.fifo",
        configureOptions: (message, options) =>
        {
            options.MessageGroupId = message.BidId;
            options.MessageDeduplicationId = message.BidId;
        });
});
```

### Other Information

Configuration supported by code configuration only.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS.Messaging (or related) package versions

AWS.Messaging

### Targeted .NET Platform

*

### Operating System and version

*

sankettangade:
Thanks for the feature request and creating a PR with implementation. Team will review the PR.

PR #341, opened 2026-05-13
Per message options configuration for `IMessagePublisher`

EtherZa:
#340 

Adds optional per-message `configureOptions` callbacks to `IMessageBusBuilder.AddSQSPublisher` and `IMessageBusBuilder.AddSNSPublisher`.

With this change, applications using the generic `IMessagePublisher` can populate SQS- and SNS-specific publish options from the message being sent. This enables scenarios such as publishing to FIFO queues and topics by deriving values like `MessageGroupId` and `MessageDeduplicationId` from the message itself, while still using type-based routing through `IMessagePublisher`.
The change also allows these callbacks to override service-configured defaults for a given publish operation, providing additional flexibility when message-specific behavior is required. Documentation has been updated to describe this capability in the publishing guidance.

Example usage:
```csharp
builder.Services.AddAWSMessageBus(builder =>
{
    builder.AddSQSPublisher<TransactionInfo>(
        "https://sqs.us-west-2.amazonaws.com/012345678910/MyFifoQueue.fifo",
        configureOptions: (message, options) =>
        {
            options.MessageGroupId = message.TransactionId;
            options.MessageDeduplicationId = message.TransactionId;
        });

    builder.AddSNSPublisher<BidInfo>(
        "arn:aws:sns:us-west-2:012345678910:MyFifoTopic.fifo",
        configureOptions: (message, options) =>
        {
            options.MessageGroupId = message.BidId;
            options.MessageDeduplicationId = message.BidId;
        });
});
```

### DIFF
--- a/.autover/changes/e46a0c0e-61c2-4c99-9dd6-173251c999ab.json
+++ b/.autover/changes/e46a0c0e-61c2-4c99-9dd6-173251c999ab.json
@@ -1,0 +1,11 @@
+{
+  "Projects": [
+    {
+      "Name": "AWS.Messaging",
+      "Type": "Minor",
+      "ChangelogMessages": [
+        "Added per-message SQS and SNS publish option configuration for IMessagePublisher, including FIFO support."
+      ]
+    }
+  ]
+}
--- a/.autover/changes/e46a0c0e-61c2-4c99-9dd6-173251c999ab.json
+++ b/.autover/changes/e46a0c0e-61c2-4c99-9dd6-173251c999ab.json
@@ -2,7 +2,7 @@
   "Projects": [
     {
       "Name": "AWS.Messaging",
-      "Type": "Minor",
+      "Type": "Major",
       "ChangelogMessages": [
         "Added per-message SQS and SNS publish option configuration for IMessagePublisher, including FIFO support."
       ]
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ builder.Services.AddAWSMessageBus(builder =>
 {
     builder.AddSQSPublisher<TransactionInfo>(
         "https://sqs.us-west-2.amazonaws.com/012345678910/MyFifoQueue.fifo",
-        configureOptions: (message, options) =>
+        configureOptions: (serviceProvider, message, options) =>
         {
             options.MessageGroupId = message.TransactionId;
             options.MessageDeduplicationId = message.TransactionId;
@@ -66,9 +66,12 @@ builder.Services.AddAWSMessageBus(builder =>
 
     builder.AddSNSPublisher<BidInfo>(
         "arn:aws:sns:us-west-2:012345678910:MyFifoTopic.fifo",
-        configureOptions: (message, options) =>
+        configureOptions: async (serviceProvider, message, options, cancellationToken) =>
         {
-            options.MessageGroupId = message.BidId;
+            var sharedState = serviceProvider.GetRequiredService<SharedState>();
+            var groupId = await sharedState.GetGroupId(message, cancellationToken);
+
+            options.MessageGroupId = groupId;
             options.MessageDeduplicationId = message.BidId;
         });
 });
@@ -559,7 +562,7 @@ In the following example the framework will check every 1 second for messages th
 {
     options.VisibilityTimeout = 30;
     options.VisibilityTimeoutExtensionThreshold = 5;
-    VisibilityTimeoutExtensionHeartbeatInterval = 1;
+    options.VisibilityTimeoutExtensionHeartbeatInterval = 1;
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,29 @@ builder.Services.AddAWSMessageBus(builder =>
 
 Once you have registered the framework during startup, inject the generic `IMessagePublisher` into your code. Call its `PublishAsync` method to publish any of the message types that were configured above. The generic publisher will determine the destination to route the message to based on its type.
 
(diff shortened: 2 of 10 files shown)


normj (commented):
I like the idea of this PR. I might be over thinking this but just wondering if a user might need access to the services registered in the DI container. For example use some DI service that would be able to look up the group id based on a customer id on the on the message. In that case the service call might be async so it might be good then for the callbacks to return ValueTask.

EtherZa:
@normj I have added per-message options configuration for publishing to the `EventBridge` too. It probably won't have too much real world usage, but it is opt-in and keeps expectations aligned.

`IServiceProvider` has been added with overloads for `Action`<...> and `Func<..., ValueTask>` (syntactic sugar to wrap Action in a Func). It has unfortunately come with some small, but breaking changes for which I have bumped the release from a minor to major (but please review and adjust accordingly).

Breaking changes:
1. `MessageRoutingPublisher (IMessagePublisher)` is no longer registered as a singleton. The `IServiceProvider` instance needs to be scoped with the calling class so that it can be passed on to the config action/func.
2. `ServiceCollectionExtensions.AddAWSMessageBus` was exposing `MessageBusBuilder` instead of the interface. I suspect that it was an oversight and has been changed to keep the overloads in the interface where possible. If you would like to avoid this change, the overloads can be moved to the `MessageBusBuilder` class.

These are both technically breaking changes, but may not warrant a full bump.

```cs
builder.Services.AddAWSMessageBus(builder =>
{
    // synchronous
    builder.AddSQSPublisher<TransactionInfo>(
        "https://sqs.us-west-2.amazonaws.com/012345678910/MyFifoQueue.fifo",
        configureOptions: (serviceProvider, message, options) =>
        {
            options.MessageGroupId = message.TransactionId;
            options.MessageDeduplicationId = message.TransactionId;
        });

    // asynchronous
    builder.AddSNSPublisher<BidInfo>(
        "arn:aws:sns:us-west-2:012345678910:MyFifoTopic.fifo",
        configureOptions: async (serviceProvider, message, options, cancellationToken) =>
        {
            var service =  serviceProvider.GetRequiredService<SomeService>();
            var messageGroupId = service.GetMessageGroupIdAsync(message, cancellationToken0;

            options.MessageGroupId = messageGroupId;
        });
});
```

I have left each commit unsquashed to make rollback, etc easier. Please modify/drop as you see fit.

EtherZa:
@normj / @GarrettBeatty any thoughts on this PR? 

It would be useful for leveraging fair queues as well.

GarrettBeatty (commented):

**Adversarial review — 7 findings (1 to verify, 1 blocking, 3 major API/binary-compat, 2 minor).** The feature (per-message `configureOptions` callbacks for SQS/SNS/EventBridge) is sound; concerns cluster around back-compat and one DI lifetime change — consistent with the `Major` bump the change file declares, but under-documented.

**1. [verify] Publisher DI registration.** `MessageRoutingPublisher.PublishAsync` now resolves `ISQSPublisher` / `ISNSPublisher` / `IEventBridgePublisher` via `GetRequiredService<>()` per publish (replacing `ActivatorUtilities.CreateInstance` + cache). Only the tests are shown registering those three interfaces. If the framework's own wiring (`ServiceCollectionExtensions` / `MessageBusBuilder.Build`) doesn't register them, every publish through the generic `IMessagePublisher` throws `InvalidOperationException` at runtime. Please confirm production wiring registers all three, and add an end-to-end test that resolves `IMessagePublisher` from a provider built **only** via `AddAWSMessageBus(...)`.

**2. [blocking] `IMessagePublisher` lifetime Singleton → Scoped.** `MessageBusBuilder.Build()` changed `TryAddSingleton` → `TryAddScoped` on `MessageRoutingPublisher`. This breaks consumers that resolve `IMessagePublisher` from the root provider or hold it in a singleton / `IHostedService`: "cannot resolve scoped service from root provider" under scope validation, or a captive-dependency bug otherwise. The change is undocumented. Please justify it (or create a scope internally per publish and keep the publisher singleton), and document the migration.

**3. [major — source + binary break] Concrete 2-arg overload removed.** On the concrete `MessageBusBuilder`, the 2-arg `AddSQSPublisher(queueUrl[, messageTypeIdentifier])` (and the SNS/EventBridge equivalents) is removed — the old `messageTypeIdentifier = null` default is gone, surviving only as an interface DIM. Callers typed to the concrete `MessageBusBuilder` stop compiling, and pre-compiled callers hit a binary break. Recommend keeping the original concrete overloads alongside the new ones.

**4. [major — source + binary break] `AddAWSMessageBus` parameter widened.** `Action<MessageBusBuilder>` → `Action<IMessageBusBuilder>`. Binary break (signature change), plus a source break for callers who declared the delegate explicitly as `Action<MessageBusBuilder>` or used members that exist only on the concrete builder. Consider adding an `Action<IMessageBusBuilder>` overload rather than replacing the existing one, and confirm the interface exposes every previously-reachable member.

**5. [major — binary break for implementers] New abstract members on public interface.** `IMessageBusBuilder` gains new **abstract** members (the `(queueUrl, messageTypeIdentifier, Action<...>)` / `Func<...>` overloads and SNS/EventBridge equivalents). The forwarding DIMs are non-breaking, but abstract additions break any external type implementing `IMessageBusBuilder` compiled against the old interface. Give each newly-required member a default implementation, or move them to the concrete class only.

**6. [minor] `<inheritdoc/>` on `PublisherMapping.ConfigureOptions`.** `PublisherMapping` is a class implementing no interface member with that doc, so `<inheritdoc/>` resolves to nothing (empty API doc / doc-build warning). Replace with a real `<summary>` describing the callback and its invocation timing.

**7. [minor] Change file omits the breaking changes.** `.autover/changes/*.json` is correctly typed `Major`, but its single `ChangelogMessage` only advertises the new feature. For a breaking release it should enumerate the breaks: `IMessagePublisher` Singleton→Scoped, the `AddAWSMessageBus` param change, and the removed concrete overloads — with migration guidance.

_Lower-confidence items were considered and dropped: options-precedence (the code passes `null` when there's no callback, so service/caller defaults are preserved), a `message!` NRE (null is guarded earlier in `PublishAsync`), `ConfigureAwait(false)`, the exception-type change, cache removal, and README example style._

Issue #286, opened 2025-09-09
Feature request: AWS.Messaging.Testing package


JanHyka:
### Describe the feature

Introduce a package with tooling that allows for easy testing with AWS.messaging flows - from having payload enqueued all the way to consumption in handler.

### Use Case

Writing fast and simple test for flows in messaging (ie from enqueued event to consumption in your handler) is possible, but not really trivial unless you have good knowledge of event structures... and requires a bit of boilerplating. Having testing toolkit should make it easier for everyone to use.

### Proposed Solution

Introduce .Testing package that allows easy to use setup to test your AWS.messaging flows without necessity to connect to real SQS.

This could be just set of methods that will help you configure aws.messaging to local mock endpoint + `EventMock< T>` sort of wrapper for adding proper envelopes... Or go all the way including wiremock or similar to mock entire SQS and just use service collection extension method to inject it for testing scenarios.

### Other Information

This is early idea that would require proper feature list (wishlist) for desired functionality before design and implementation.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS.Messaging (or related) package versions

AWS.Messaging 1.0.0+

### Targeted .NET Platform

NET6+

### Operating System and version

any

slang25:
This would be really cool.

Just an option for you if you are using SNS & SQS and want something that works in memory, I wrote a library that mocks those services, and try to pair to to the AWS behaviour as close as possible for general purpose messaging scenarios:
https://github.com/justeattakeaway/LocalSqsSnsMessaging

PR #344, opened 2026-05-14
Introduce TimeProvider in place of DateTimeHandler

JanHyka:
*Issue #, if available:*

#286 

*Description of changes:*

Initial step allowing implementation of #286.

This pull request updates the AWS.Messaging library to replace the custom `DateTimeHandler` abstraction with the .NET 8 `TimeProvider` for handling time-related functionality. This change modernizes the codebase, improves testability, and removes now-unnecessary code. Several classes and their constructors have been updated to use `TimeProvider`, and related tests and dependency registrations have also been adjusted.

Key changes include:

**Migration to .NET 8 TimeProvider:**

* Replaced all usages of `IDateTimeHandler` and `DateTimeHandler` with the built-in `TimeProvider` throughout the codebase, including in `EnvelopeSerializer`, `SingleTypeSqsPollerEnvelopeSerializer`, `BackoffHandler`, `DefaultMessageManager`, and `InFlightMetadata`. 

**Removal of Obsolete Code:**

* Deleted the now-unnecessary `DateTimeHandler` and `IDateTimeHandler` files and interfaces.
* Constructors of public API classes missing TimeProvider internally use `TimeProvider.System` and are marked as `[Obsolete]` - this is to reduce risk of unexpected behavior in case custom TimeProvider is used and propagation chain gets silently broken.

**Dependency Injection and Constructor Updates:**

* Updated dependency injection registrations and factory methods to provide `TimeProvider.System` where required. Added new constructor overloads and marked old ones as obsolete to ease migration.

**Test Adjustments and Improvements:**

* Updated unit tests to use the new `TimeProvider` abstraction, including using `FakeTimeProvider` for improved testability. Added the `Microsoft.Extensions.TimeProvider.Testing` NuGet package to test projects. 
* Work around thread safety issue in TestLambdaLogger, allowing significant unit test speedup
* Unit test accumulative duration drop to ~50% (1.2min -> 38.4s)   

**Changelog and Documentation:**

* Added a changelog entry indicating the migration from `DateTimeHandler` to .NET 8 `TimeProvider`.


GarrettBeatty (commented):

**Verdict: 7 findings — 3 blocker, 3 major, 1 minor.** The `TimeProvider` migration is sound in shape, but there's a potential 1000× timing regression and two public-API breaks mislabeled as a Patch, plus a flaky/weak rewritten test.

1. **[blocker]** `src/AWS.Messaging/Services/DefaultMessageManager.cs` — `ExtendUnfinishedMessageVisibilityTimeoutBatch`. The old code passed `_configuration.VisibilityTimeoutExtensionHeartbeatInterval` bare to `Task.Delay(...)` (int → **milliseconds**); the new code wraps it in `TimeSpan.FromSeconds(...)` (**seconds**). If that property is an `int`, this is a silent **1000× behavior change** on the live SQS visibility-extension loop. The sibling `VisibilityTimeout` is used as `TimeSpan.FromSeconds(...)`, which suggests seconds was intended all along (making the old code a latent bug and this a fix) — but please confirm against the property's declared unit and call it out in the changelog either way.

2. **[blocker]** `src/AWS.Messaging/Services/IDateTimeHandler.cs` — a **public** interface (`public interface IDateTimeHandler`) is deleted. Removing an exported public type is a breaking API change; it can't ship as a `Patch`, and the migration isn't disclosed in the changelog message.

3. **[blocker]** `src/AWS.Messaging/Serialization/EnvelopeSerializer.cs` (ctor) — the **public** constructor's signature was changed in place (`IDateTimeHandler` → `TimeProvider`) with **no** `[Obsolete]` overload retained, unlike `DefaultMessageManager` which correctly kept its old ctor. Either retain an obsolete overload defaulting to `TimeProvider.System`, or confirm the type isn't meant to be externally constructed and document the break.

4. **[major]** `.autover/changes/…json` — `Type` is `"Patch"`, but findings #2/#3 make this at least a minor/breaking change. Fix the AutoVer type so versioning warns consumers.

5. **[major]** `test/AWS.Messaging.UnitTests/Backoff/BackoffHandlerTests.cs` — `RetryAsync_IntervalBackoff` timing race. The rewrite runs `BackoffAsync` on a `Task.Run` background task and relies on a real `await Task.Delay(10)` to *hope* the task has parked in `Task.Delay` before calling `fakeTimeProvider.Advance(1s)`. There's no synchronization guaranteeing the timer is registered before the advance — under CI load this can hang or under-count. Gate the advance on the timer actually being scheduled (e.g. a `TaskCompletionSource`).

6. **[major]** `test/AWS.Messaging.UnitTests/Backoff/BackoffHandlerTests.cs` — same test, assertion weakened to `Times.AtMost(retries)`, which **passes even on zero invocations**. Combined with the race in #5, the test can be green while proving nothing. Use `Times.Exactly(retries)`.

7. **[major]** `src/AWS.Messaging/Services/Backoff/BackoffHandler.cs` (ctor) — gained a required `TimeProvider` param with no obsolete overload. `internal`, so no external break, but confirm all construction sites are updated and that DI actually resolves `TimeProvider` for it at runtime.

8. **[minor]** `src/AWS.Messaging/Serialization/EnvelopeSerializer.cs` / `SingleTypeSqsPollerEnvelopeSerializer.cs` (ctors) — add `ArgumentNullException.ThrowIfNull(timeProvider)` on the changed constructors so a null clock fails at construction rather than as an NRE at the first `GetUtcNow()`.

